PR #1, opened 2020-04-29
Rewrite build on Kotlin DSL.

ileasile:
Build script was rewritten on Kotlin DSL.
Kotlin version was updated to 1.3.71 (last stable one).
Also, moved Kotlin plugin version to settings.gradle.kts to avoid explicitly mentioning it everywhere (in root project and maybe in subprojects which may appear)


ileasile:
Closing it because of growing number of differences and no need in transfering build process to Kotlin DSL right now.

PR #10, opened 2020-10-28
First batch of tutorials

MKhalusova:
While I continue to work on transfer learning tutorial and Keras model import, take a look at the first batch of tutorials. These are about training a model from scratch and using it for Inference. I tried to make the tutorials as simple/short/beginner-friendly as possible.


docs/training_a_model.md
```diff
@@ -0,0 +1,90 @@
...
+
+```
+Accuracy: 0.8919999599456787
+```
+
+If you are happy with the model's evaluation metric, you can save the model for future use in your production environment.  
+
+```kotlin
+model.save(File("src/model/my_model"))
+model.close()
```
zaleslaw, line 86: Please format code in tutorial to use 'use' clause for AutoClosebale objects like a model (it will be more Kotlin-idiomatic)

Also, add a note about that model should be closed (manually or autoclosed with 'use' clause)

docs/training_a_model.md
```diff
@@ -0,0 +1,90 @@
...
+Once the model has been trained, it's important to evaluate its performance on the validation dataset, so that you can 
+check how it generalizes to the new data. 
+
+```kotlin
+val accuracy = model.evaluate(dataset = validation, batchSize = 100).metrics[Metrics.ACCURACY]
+println("Accuracy: $accuracy")
+```
+
+```
+Accuracy: 0.8919999599456787
```
zaleslaw, line 79: Maybe add a note, that results are not determenistic and could be different on different machines and so on

docs/quick_start_guide.md
```diff
@@ -0,0 +1,63 @@
+# Quick Start Guide
+
+Whether you want to add Kotlin DL into your project, or experiment with it in a Jupyter Notebook, first you'll need to 
+set up appropriate dependencies.  
+
+## Working with Kotlin DL in an IDE
+1. Open a Kotlin project where you want to use Kotlin DL, or create a new Kotlin project in IntelliJ IDEA as described 
+in [Kotlin documentation](https://kotlinlang.org/docs/tutorials/jvm-get-started.html).
+2. Add the Kotlin DL dependency to your project's build file. 
+    * If you're using Gradle as build system, add the following to the `build.gradle.kts` file:
```
zaleslaw, line 10: Could you please add the instruction for pure Gradle (not kts) too

docs/create_your_first_nn.md
```diff
@@ -0,0 +1,118 @@
...
+The basic building block of a neural network is a **layer**. To define a neural network, you need to describe 
+what layers it should consist of. The goal of those layers is to capture some data representation aspects during 
+the training phase. With enough data, layers, and training time, neural networks are capable of capturing enough 
+complexity to result in great performance on many tasks. 
+
+Here, our neural network will consist of a few simple layers in a sequence:  
+
+```kotlin
+private val model = Sequential.of(
+    Input(28,28),
```
zaleslaw, line 44: Looks like it should be Input (28, 28, 1).  We should be careful about input data shapes

zaleslaw: 1 - is number of channels (1 for greyscale or 3 for typical png image)

Maybe I need to add a warning about that

docs/create_your_first_nn.md
```diff
@@ -0,0 +1,118 @@
...
+ 
+Next, we've got a sequence of three **Dense()** layers. Dense layers are the most common layers in all kinds
+ of neural network architectures. They are also called fully connected layers because they connect every neuron 
+ in one layer to every neuron in the next layer.
+ 
+For each Dense layer we have specified its size. The first one has 300 neurons (or nodes), the second one has 100 neurons, 
+and the last one has 10 neurons (remember, we have 10 different classes to predict). 
+
+In each Dense layer we have also explicitly specified an activation function to be used. 
+Without activation functions a neural network would simply be passing data through linear nodes and would not be performing
```
zaleslaw, line 67: public val activation: Activations = Activations.Relu,

Dense and Conv2D layers use Relu by default. 

Defaults are different from Keras

docs/create_your_first_nn.md
```diff
@@ -0,0 +1,118 @@
...
+    model.summary()
+```
+
+```
+INFO  api.core.Sequential - =================================================================
+INFO  api.core.Sequential - Model: Sequential
+INFO  api.core.Sequential - _________________________________________________________________
+INFO  api.core.Sequential - Layer (type)                 Output Shape              Param #   
+INFO  api.core.Sequential - =================================================================
+INFO  api.core.Sequential - flatten_1(Flatten)           [784]                     0
```
zaleslaw, line 100: It will have alternative output in new artifact (output shapes will be changed)

docs/loading_trained_model_for_inference.md
```diff
@@ -0,0 +1,49 @@
...
+![](images/test-image-bag.png)
+
+The model expects as an input a 28 x 28 Tensor (Tensor == multi-dimensional array). We can read in the image as 
+a FloatArray, and define a function that will help us reshape a regular FloatArray into a Tensor with appropriate 
+dimensions:
+
+```kotlin
+val floatArray = ImageConverter.toRawFloatArray(File(PATH_TO_TEST_IMAGE))
+
+private fun reshape(floats: FloatArray): Tensor<*> {
```
zaleslaw, line 22: hmmm, seems like only one place where Tensors from low-level API are accessible, need to refactor and update later

I suppose that reshape function should return FloatArray (reshaped variable) instead of Tensor<*>

docs/quick_start_guide.md
```diff
@@ -0,0 +1,63 @@
...
+     <groupId>org.jetbrains.kotlin.kotlin-dl</groupId>
+     <artifactId>api</artifactId>
+     <version>[Kotlin DL version]</version>
+     <type>pom</type>
+   </dependency>
+   ```
+ 
+You can also take advantage of Kotlin DL functionality in any existing Java project, even if you have no other Kotlin 
+code in it yet. Check out [these instructions](https://kotlinlang.org/docs/tutorials/mixing-java-kotlin-intellij.html#adding-kotlin-source-code-to-an-existing-java-project) 
+on how to add Kotlin code  in your existing Java code base.
```
zaleslaw, line 40: I have an example of simple LeNet in Java https://github.com/JetBrains/KotlinDL/blob/release_0.1/examples/src/main/java/LeNetClassic.java
maybe it will be useful to add

zaleslaw (changes requested):
Great tutorials! Add more!

docs/quick_start_guide.md
MKhalusova: Perhaps a stupid question, but is it going to be different?

docs/create_your_first_nn.md
MKhalusova: When will the new artefact be available so I could check that?

docs/loading_trained_model_for_inference.md
MKhalusova: Another thought that I had here - should the reshaping happen within loaded model? Or maybe it should be part of the preprocessing before the data gets fed into the model?

docs/quick_start_guide.md
MKhalusova: That's not what I meant by adding Kotlin code to an existing Java code base. I mean having .kt files in a Java code base.

docs/training_a_model.md
MKhalusova: Done

MKhalusova:
@zaleslaw I've added your suggestions. Please take another look. The rest of the tutorials will be in a different pull request, so if you're happy with these, feel free to merge them into the release branch.

PR #26, opened 2020-12-19
`ZeroPadding2D` layer implementation

AntKos:
This is the implementation of `ZeroPadding2D` layer. The usage of this layer can be found at tensorflow implementation of Resnet (see https://github.com/tensorflow/tensorflow/blob/582c8d236cb079023657287c318ff26adb239002/tensorflow/python/keras/applications/resnet.py#L175).

This layer has the following structure in `.json` configs:
`{
        "class_name": "ZeroPadding2D",
        "config": {
          "name": "conv1_pad",
          "trainable": true,
          "dtype": "float32",
          "padding": [[3,3], [3,3]],
          "data_format": "channels_last"
}`
As its `config.padding` has different type than `Convolution`/`Pooling` layers, I suggest to use `KerasPadding` class instead of plain `String` in `LayerConfig`. A custom json converter class `PaddingConverter` is used to maintain different `padding` set-ups.

### DIFF
--- a/api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/twodim/ZeroPadding2D.kt
+++ b/api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/twodim/ZeroPadding2D.kt
@@ -13,34 +13,91 @@ import org.tensorflow.Operand
 import org.tensorflow.Shape
 import org.tensorflow.op.Ops
 
-class ZeroPadding2D(
-    val padding: IntArray,
-    private val dataFormat: String? = CHANNELS_LAST,
-    name: String = ""
-) : Layer(name) {
+/**
+ * Zero-padding layer, which adds zeros on sides of image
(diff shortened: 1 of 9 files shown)


zaleslaw:
First of all, dear @AntKos thanks for the contirbution and your thoughts about ResNet implementation.

It looks enough good for review, but could I suggest you to add some documentation for public classes/methods and main hacks (if it's used)?

About plans for this PR: after review (it takes some time due to holidays) it could be merged after releasing 0.1.1 bug fix release (no big API changes there).

P.S. The main stopper here is Merge layer (Add, for example) based on ConcatV2 operand in TF.
Also, I'm working on BatchNorm layer, another big pain here. 

Feel free to share your thoughts and code experiments here or in slack-channel #deeplearning in [Kotlin slack](https://kotlinlang.slack.com/)

AntKos:
@zaleslaw Thank you for such warm welcome and explicit explanation of future plans and blocks! I added a bit of javadocs, and also fixed a bug with saving model caused by missing `Converter`.

zaleslaw:
@AntKos Thanks for update, it looks better now. I return with detailed review in January, welcome to contribute any functionality, new layers and etc!

Issue #23, opened 2020-12-19
[Feature Request] Support for functional API


Rishit-dagli:
As of now, only Sequential models are supported it would be great to allow models to take inspiration from [Keras Functional APIs](https://www.tensorflow.org/guide/keras/functional) and allow building such models. This would allow creating models that are more flexible with KotlinDL and also open up more possibilities with the kind of models one could make with KotlinDL including branching, multiple-inputs, multiple-outputs and so on

zaleslaw:
Hi, agree, this is an important feature. 
Currently, it's not supported due to limitation of underlying Java API for TF 1.15 and limitation of C/C++ API:

- Functional API totally based on Concatenate and its children usage (Add, Subtract and another Merge layers)
- Concat operand is available, but gradients are not available for it
- no ability to register custom gradient for analogue of Concat op
- Java TF API for 1.x is frozen

**Possible solution:**

- Move to Java TF API 2.x (stable version will be released in 2021)
- Update Java TF API with custom gradients for Concat Op (because it has the same [issue](https://github.com/tensorflow/java/issues/156)
- Write Concatenate-based layers family
- Provide Functional API

As you see, there a lot of work, which should be done in 2021 with TF Java API community

zaleslaw:
So, the early version of Functional API is merged. I close this issue

PR #42, opened 2021-03-15
Early version of ModelLoading

zaleslaw:


zaleslaw:
The early draft of Functional API is provided here. Related to #23

PR #65, opened 2021-05-20
Update README.md

apahl:
The original listed dependency did not work for me. Using the one from the `build.gradle` example does work.


zaleslaw:
Many thanks for your fix

apahl:
Thank you for this very exciting library!

PR #66, opened 2021-05-21
Add parametrized ELU layer with tests

### LINKED ISSUE #55: Add ELU layer
We are missing some activation layers to support the export of models from Keras fully. One of them is the ELU layer.

Add an activation layer class, write documentation for it, write a test for it, try, if possible, create a small trainable network with it (in your own GitHub) and attach a link here in the comments.

The layer should be placed [here](https://github.com/JetBrains/KotlinDL/tree/master/api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/activation)

As a reference implementation, the [ReLU activation layer](https://github.com/JetBrains/KotlinDL/blob/master/api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/activation/ReLU.kt) could be used, but feel free to improve it!

Also, support for export and import of activation layer in JSON format should be added (see [ModelLoader.kt ](https://github.com/JetBrains/KotlinDL/blob/master/api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/inference/keras/ModelLoader.kt) and [ModelSaver.kt](https://github.com/JetBrains/KotlinDL/blob/master/api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/inference/keras/ModelSaver.kt))

A detailed description of the activation layer can be found [here](https://keras.io/api/layers/activation_layers/elu/)

avan1235:
Resolves #55 

I found again the org.jetbrains.kotlinx.dl.api.core.layer.core package with many default activation function implementations. Is it going to be removed and replaced by single class with documentation for every activation layer?


zaleslaw:
@avan1235 Regarding your question: we are working on full compatibility with Keras now, so Keras have a separate [Activation layer](https://keras.io/api/layers/core_layers/activation/) and [activations layers](https://keras.io/api/layers/activation_layers/) with some extended functionality (like in the ReLU case).

At this moment I suggest having the full copy of such layers for easy mapping but keep in mind (after double-checking) that in the future these layers on the KotlinDL side could be merged together (in one Activation layer or something else)

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/activation/Activation.kt
```diff
@@ -0,0 +1,37 @@
...
+
+package org.jetbrains.kotlinx.dl.api.core.layer.activation
+
+import org.jetbrains.kotlinx.dl.api.core.KGraph
+import org.jetbrains.kotlinx.dl.api.core.layer.Layer
+import org.tensorflow.Operand
+import org.tensorflow.Shape
+import org.tensorflow.op.Ops
+
+public abstract class Activation(name: String) : Layer(name) {
```
zaleslaw, line 14: Haha :), we have yet another "Activation" class, maybe it better to rename it to AbstractActivationLayer?
Maybe "core" ActivationLayer could be inherited from this too? What do you think?

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/activation/ELU.kt
```diff
@@ -0,0 +1,51 @@
...
+ * the activation closer to zero which enable faster learning as they
+ * bring the gradient to the natural gradient.
+ *
+ * @property [alpha] Hyperparameter that controls the value to which
+ * an ELU saturates for negative net inputs. Should be > 0.
+ * @constructor Creates [ELU] object.
+ * @since 0.3
+ */
+public class ELU(
+    public val alpha: Float = 1.0f,
```
zaleslaw, line 30: looks like this parameter is the reason for the separate ELU layer in Keras (like for ReLU too) because the simplest activation usually have no parameters

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/activation/ELU.kt
```diff
@@ -0,0 +1,51 @@
...
+
+    override fun forward(
+        tf: Ops,
+        input: Operand<Float>
+    ): Operand<Float> = when(alpha) {
+        1.0f -> tf.nn.elu(input)
+        else -> {
+            val greaterThanZero = tf.math.greater(input, tf.constant(0.0f))
+            val scaledActivation = tf.math.mul(tf.constant(alpha), tf.nn.elu(input))
+            tf.where3(greaterThanZero, input, scaledActivation)
```
zaleslaw, line 45: do we have a chance to avoid where op here (I'm not sure, but maybe the gradients are missed for this and this layer should be marked with NoGradients interface and should be frozen by default) Don't mind if you have no able to rewrite it, leave it this way, but mark with NoGradient interface

api/src/test/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/ReLUTest.kt
```diff
@@ -10,36 +10,27 @@ import org.junit.jupiter.api.Assertions
...
     @Test
     fun reluWithNegativeSlope() {
-        val input = floatArrayOf(-3.0f, -1.0f, 0.0f, 2.0f)
-        val expected = floatArrayOf(-3.0f, -1.0f, 0.0f, 2.0f)
-        val actual = floatArrayOf(0.0f, 0.0f, 0.0f, 0.0f)
-
-
         val exception = Assertions.assertThrows(UnsupportedOperationException::class.java) {
-            assertActivationFunction(ReLU(negativeSlope = 1.0f), input, actual, expected)
+            assertActivationFunctionIrrelevantInputOutput(ReLU(negativeSlope = 1.0f))
```
zaleslaw, line 33: Remark: I think it's a good idea for tests, we need more special testing functions to reuse test-scenarios

zaleslaw (changes requested):
So, I'll run the TC for this PR and write about the results
We will merge it this week

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/activation/ELU.kt
avan1235: that convinces me that it could be somehow "duplicated code" but in fact with other features and use cases

api/src/test/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/ReLUTest.kt
avan1235: will try to introduce some if see any possibility to abstract here 👍🏻

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/activation/Activation.kt
avan1235: great observation, will fix this in this way

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/activation/ELU.kt
avan1235: imo we can mark the AbstractActivationLayer as not trainable and so all they child classes as it's seems to be more common for the activations not to be trainable (and as I correctly understand it will not be necessary to use the NoGradient interface then and even harder to override in some subclasses which could have been trainable).

Btw interface here doesn't look like the best approach, maybe it should be redesigned with annotation (so it could mark only specific class and not its children)?

zaleslaw: At this time, NoGradient is a marker interface, and it's very simple to verification. There is no need to mark the AbstractActivationLayer with this interface because it's not true for the ReLU layer, which could participate in backward computation.

It looks like NoGradient is an unclear name for the interface. Still, the idea is related to weights (activation is just a function applied to the previous layer's outputs), but the problem is in having the GradientOp for this TF Op. For many TF Ops gradients (read here: a building block for backward computing is missed).

I suggest don't concentrate on this now, but just marked the class with this interface. Later I'll do a double-check for used TF ops

zaleslaw:
LGTM for me, will be merged after running TC on my side

zaleslaw (approved):

Issue #52, opened 2021-05-18
Add Softmax layer


zaleslaw:
We are missing some activation layers to support the export of models from Keras fully. One of them is the Softmax layer.

Add an activation layer class, write documentation for it, write a test for it, try, if possible, create a small trainable network with it (in your own GitHub) and attach a link here in the comments.

The layer should be placed [here](https://github.com/JetBrains/KotlinDL/tree/master/api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/activation)

As a reference implementation, the [ReLU activation layer](https://github.com/JetBrains/KotlinDL/blob/master/api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/activation/ReLU.kt) could be used, but feel free to improve it!

Also, support for export and import of activation layer in JSON format should be added (see [ModelLoader.kt ](https://github.com/JetBrains/KotlinDL/blob/master/api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/inference/keras/ModelLoader.kt) and [ModelSaver.kt](https://github.com/JetBrains/KotlinDL/blob/master/api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/inference/keras/ModelSaver.kt))

A detailed description of the activation layer can be found [here](https://keras.io/api/layers/activation_layers/softmax/)

d-lowl:
I'm currently looking into this issue and have a question regarded the expected implementation. TF/Keras Softmax layer in Python takes two arguments: the input tensor and the mask. However, the current abstract forward function does not allow that. Do we want to implement masking (and hence extending the signature for `forward`) or is it out of scope for this issue?

zaleslaw:
@d-lowl It's a reasonable question: so, I suppose we can implement this layer without a mask.

As you rightly noted, at the moment there is no support for masks, including at the level of layers and arguments. The fact is that while they are not needed for those models that are supported in the model zoo. I have a secret hope that they may not be needed.

d-lowl:
Gotcha. I'll hopefully come back with a PR soon.

zaleslaw:
So, I've assign it for you, if you are agree with that, to avoid two or more PRs for one tickets

PR #76, opened 2021-05-30
Softmax layer

d-lowl:
Adds Softmax layer implementation and tests (#52).

Example repo (instructions included): https://github.com/d-lowl/fantastic-bassoon

Limitations:
* As discussed with zaleslaw, masks are not implemented, as they are out of scope for this issue. To be implemented later.
* ~~Only the last axis is supported for normilisation (i.e. the default axis). The reason -- Tensorflow Java lib's softmax implementation does not have axis parameter (as opposed to Python's implementation). If not needed at this point, I can remove the WIP and the PR can be merged. Otherwise, I'm working on it now.~~ See update


d-lowl:
UPD: I've implemented handling normalisation along non-last axis, essentially back-porting this functionality from Tensorflow 2 (it's not available in 1.15 API). Should be good for review now.

For tests I've added another function for asserting on 2D inputs to activation layers. I was trying to generalise with the existing functions (by avan1235), but converting from Tensors/Operands to Array<*> conflicts with doing the same for FloatArrays. So I gave up on that.

zaleslaw:
So, thanks for the PR and great refactoring of ModelSaver, please wait a few days with this PR before merging, there is a queue of PRs with changes in ModelSaver

Hope you will be closing this chain of PRs related to the activation layers and will make the same refactoring in the "when" op in ModelSaver? 

Also, I see a great example project with a demonstration of the Softmax layer, maybe later we will add it as an example

d-lowl:
Yeah, cheers. 

> please wait a few days 
of course, no rush here

> and will make the same refactoring in the "when" op in ModelSaver?
If I remember correctly there are a couple more places where `when` clause can be switched from "class patterns" to "is pattern". But I haven't touched them, as unrelated to that particular layer. I guess they can also be switched.

zaleslaw:
So, I agree with you, this kind of refactoring is out of scope but will be useful later

zaleslaw:
So, @d-lowl it's your turn! Please merge the master, fix some refactoring issue and I'll run the TC after that

d-lowl:
Merged, when clauses are refactored (in the last commit), good to go from my side 👌

zaleslaw:
Great PR @d-lowl thanks for the contrib!

Issue #51, opened 2021-05-18
Add Orthogonal initializer


zaleslaw:
We are missing some initializers to fully support the export of models from Keras. One of them is the Orthogonal initializer.

Add an initializer class, write documentation for it, write a test for it, try, if possible, create a small trainable network with it (in your own GitHub) and attach a link.

Also, support for export and import of initializer in JSON format should be added (see [ModelLoader.kt ](https://github.com/JetBrains/KotlinDL/blob/master/api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/inference/keras/ModelLoader.kt) and [ModelSaver.kt](https://github.com/JetBrains/KotlinDL/blob/master/api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/inference/keras/ModelSaver.kt))

A detailed description of the initializer can be found [here](https://keras.io/api/layers/initializers/)

zaleslaw:
@therealansh, please leave a comment here, to assign this issue on you

therealansh:
Oh sorry @zaleslaw ! Might have missed on this one. I'll be completing the tests soon and finalising the PR.

zaleslaw:
Fixed in #84

PR #84, opened 2021-05-31
Add Orthogonal Initializer

therealansh:
This PR adds Orthogonal Initializer (#51). 

TODO:

- [x] Add Initializer implementation
- [x] Add support for export and import of layer in JSON format
- [x] Add Documentation
- [x] Add Tests

### DIFF
--- a/api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/initializer/Orthogonal.kt
+++ b/api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/initializer/Orthogonal.kt
@@ -1,0 +1,46 @@
+package org.jetbrains.kotlinx.dl.api.core.initializer
+
+import org.jetbrains.kotlinx.dl.api.core.shape.shapeOperand
+import org.jetbrains.kotlinx.dl.api.core.util.getDType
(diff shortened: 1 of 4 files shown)


therealansh:
Hey @zaleslaw , How are we writing tests for initialisers that are using RandomGenerator in it. Like using a statelessRandom will generate different Operand everytime. How can we "expect" something out of randomness?

zaleslaw:
It's a good question, it's very difficult to write good tests in ML/DL area, especially unit tests for calculations in TensorFlow due to the known "non-determinism" problem.

So, I don't know why, but "stateless"Random operand in TF has seed and it produces stable results from run to run, please have a look at the RandomNormalTest and RandomNormal internals.

So the recipe: to use seed if it's available and hope that it will be stable:(

zaleslaw:
Hi, @therealansh did you get test data from the Keras framework?
Could you please merge with the master branch, I'll run the TC builds for it

therealansh:
> Hi, @therealansh did you get test data from the Keras framework?
> Could you please merge with the master branch, I'll run the TC builds for it

Yes i took the data from tf.keras.initializers.Orthogonal

PR #97, opened 2021-06-03
Add AvgPool1D layer

### LINKED ISSUE #61: [Audio] Add AveragePooling1D layer
We are missing some layers to support the export of models from Keras fully. One of them is the AveragePooling1D layer.

Add a layer class, write documentation for it, write a test for it, try, if possible, create a small trainable network with it (in your own GitHub) and attach a link here in the comments.

The layer should be placed [here](https://github.com/JetBrains/KotlinDL/tree/master/api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/pooling)

As a reference implementation, the [AvgPool2D layer](https://github.com/JetBrains/KotlinDL/blob/master/api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/pooling/AvgPool2D.kt) could be used, but feel free to improve it!

If some refactoring to pooling layers could be applied, please, try to do it!

Also, support for export and import of layer in JSON format should be added (see [ModelLoader.kt ](https://github.com/JetBrains/KotlinDL/blob/master/api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/inference/keras/ModelLoader.kt) and [ModelSaver.kt](https://github.com/JetBrains/KotlinDL/blob/master/api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/inference/keras/ModelSaver.kt))

A detailed description of the layer can be found [here](https://keras.io/api/layers/pooling_layers/average_pooling1d/)

P.S. There are no-ops for _tf.nn.avgPool1d_ in Java API, so you could try to implement it from scratch via available ops or use tf.nn.avgPool like in TensorFlowJS with reshape in [tfjs-core/src/ops/avg_pool.ts](https://github.com/tensorflow/tfjs/blob/master/tfjs-core/src/ops/avg_pool.ts)

mkaze:
Resolves #61. This PR adds support for AvgPool1D layer.

- [x] Implement layer class
- [x] Add unit tests for layer implementation
- [x] Add support for importing models containing this layer from JSON config files
- [x] Add support for exporting models containing this layer to JSON config files


api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/pooling/AvgPool1D.kt
```diff
@@ -0,0 +1,95 @@
...
+ * @property [poolSize] Size of the temporal pooling window.
+ * @property [strides] The amount of shift for pooling window in each pooling step. If
+ * `null`, it will default to [poolSize].
+ * @property [padding] Padding strategy; can be either of [ConvPadding.VALID] which means no
+ * padding, or [ConvPadding.SAME] which means padding the input equally such that the output
+ * has the same dimension as the input.
+ * @property [dataFormat] Data format of input; can be either of [CHANNELS_LAST] or [CHANNELS_FIRST].
+ */
+public class AvgPool1D(
+    public val poolSize: Int = 2,
```
zaleslaw, line 33: I suggest turning it to 3d array of longs as we discussed earlier

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/pooling/AvgPool1D.kt
```diff
@@ -0,0 +1,95 @@
...
+    public val strides: Int? = null,
```
zaleslaw, line 34: Could we add here the default value?

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/pooling/AvgPool1D.kt
```diff
@@ -0,0 +1,95 @@
...
+    public val dataFormat: String = CHANNELS_LAST,
+    name: String = ""
+) : Layer(name) {
+
+    override val hasActivation: Boolean
+        get() = false
+    override val paramCount: Int
+        get() = 0
+    override val weights: Map<String, Array<*>>
+        get() = emptyMap()
```
zaleslaw, line 45: Please, merge with master and add support for the setter method

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/pooling/AvgPool1D.kt
```diff
@@ -0,0 +1,95 @@
...
+
+    init {
+        require(dataFormat == CHANNELS_LAST || dataFormat == CHANNELS_FIRST) {
```
zaleslaw, line 48: remove this condition

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/pooling/AvgPool1D.kt
```diff
@@ -0,0 +1,95 @@
...
+        }
+    }
+
+    override fun build(tf: Ops, kGraph: KGraph, inputShape: Shape) {}
+
+    override fun computeOutputShape(inputShape: Shape): Shape {
+        var steps = if(dataFormat == CHANNELS_LAST) inputShape.size(1) else inputShape.size(2)
+        val strideValue = strides ?: poolSize
+        steps = convOutputLength(steps, poolSize, padding, strideValue)
+        return if (dataFormat == CHANNELS_LAST) {
```
zaleslaw, line 63: remove data format

zaleslaw (changes requested):
Remove dataFormat and change the params in constructor

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/pooling/AvgPool1D.kt
mkaze: Done!

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/pooling/AvgPool1D.kt
mkaze: Done!

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/pooling/AvgPool1D.kt
mkaze: Since in pooling layers it's very common for strides to be the same as pool size value, the benefit of using `null` as default value is that the user won't need to set it explicitly and we just default to pool size value instead. This is the same behavior in Keras and PyTorch.

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/pooling/AvgPool1D.kt
mkaze: Done!

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/pooling/AvgPool1D.kt
mkaze: Done!

mkaze:
@zaleslaw I just merged it with master branch and updated it with recent changes.

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/pooling/AvgPool1D.kt
zaleslaw: Kotlin and Python users have a different experience with nullability, and Kotlin is trying to be a place without nulls.
As I commented in the MaxPool1D layer PR, please change it to the default value 1,2,1.

We need to keep control of the parameters that could be a hyper-parameter for AutoML tasks and show it to the user.
As a variant - to make a warning if stride and pool are not equal with reminder that it's a common practice

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/pooling/AvgPool1D.kt
```diff
@@ -0,0 +1,93 @@
...
+ * @property [poolSize] Size of the temporal pooling window for each dimension of input.
+ * @property [strides] The amount of shift for pooling window in each pooling step. If
+ * `null`, it will default to [poolSize].
+ * @property [padding] Padding strategy; can be either of [ConvPadding.VALID] which means no
+ * padding, or [ConvPadding.SAME] which means padding the input equally such that the output
+ * has the same dimension as the input.
+ */
+public class AvgPool1D(
+    public val poolSize: LongArray = longArrayOf(1, 2, 1),
+    public val strides: LongArray? = null,
```
zaleslaw, line 31: please convert LongArray? type to LongArray

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/pooling/AvgPool1D.kt
```diff
@@ -0,0 +1,93 @@
...
+    override var weights: Map<String, Array<*>>
+        get() = emptyMap()
+        set(value) = assignWeights(value)
+
+    init {
+        require(poolSize.size == 3) {
+            "The poolSize should be an array of size 3."
+        }
+
+        require(strides == null || strides.size == 3) {
```
zaleslaw, line 49: please change the condition

zaleslaw (changes requested):

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/pooling/AvgPool1D.kt
mkaze: I see. I understand your point about removing nullability and hidden dependencies. However, I don't get the part about losing control of the parameter during hyper-parameter tuning scenarios or architecture search. Anyways, since I am new to Kotlin I think you have a much better understanding of what would be the best decision here and I would follow it gladly :)

mkaze:
@zaleslaw I changed the strides type and default value per your request.

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/pooling/AvgPool1D.kt
zaleslaw: I meant that we train the user to use only one parameter and leave the second **null** to change together. Perhaps, looking for a new architecture, a method a la **NAS**, something with this approach will be missed, at least I think so.

PR #103, opened 2021-06-06
Improve GlobalAvgPool layers implementation and tests

mkaze:
This PR improves GlobalAvgPool layers in the following ways:

- Adds support for "data format" in `GlobalAvgPool1D`
- Adds support for "data format" in `GlobalAvgPool2D`
- Adds new test cases
- Refactors tests and makes some new abstractions for easier testing of layers
    - Specifically, one important addition is a new class, `LayerTest`, which contains a method, `runLayerInEagerMode`, which handles building and running a layer on a given input array. So every layer's test case which needs this functionality can easily use this method by subclassing the `LayerTest` class.

@zaleslaw If the changes in this PR sounds good to you, then it would be great if you could merge this PR first so that I can revise my other related pooling PRs based on the changes introduced in this PR. Thanks!


zaleslaw:
@mkaze I'd like these test improvements and LayerTest class, but I suggest postponing the channel order feature (channel last/channel first) for later.

If it is decided (after discussion in the community) to implement in the future, it should be a complex solution for all layers, probably implemented via Enum.

Highly likely, we need to think about it later if PyTorch runtime will be integrated.

Could you please cut this functionality from your PR?

mkaze:
> but I suggest postponing the channel order feature (channel last/channel first) for later.

@zaleslaw Oops! I have already implemented that in all of my other PRs (and planned to add it for convolution layers as well)... and honestly it's not that complicated to implement. But if you think it should not be added yet, then it's all right with me and I'll remove it from all of my PRs.

mkaze:
@zaleslaw I reverted the changes regarding data format.

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/pooling/GlobalAvgPool1D.kt
```diff
@@ -14,26 +14,27 @@ import org.tensorflow.op.Ops
 
 /**
  * Global average pooling operation for temporal data.
- * NOTE: Works with tensors which must have rank 3 (batch, steps, features).
- * Input shape: 3D tensor with shape `(batch_size, steps, features)`.
- * Output shape: 2D tensor with shape `(batch_size, features)`.
- * @property [name] Custom layer name.
+ *
+ * NOTE: Only works with tensors which have rank 3, i.e. tensors with shape
+ * `(batch, steps, features)`.
```
zaleslaw, line 19: Sorry, why did you remove the information about the outputSHape

api/src/test/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/GlobalAvgPooling1DTest.kt
```diff
@@ -1,13 +0,0 @@
-package org.jetbrains.kotlinx.dl.api.core.layer
-
-import org.jetbrains.kotlinx.dl.api.core.layer.pooling.GlobalAvgPool1D
-import org.junit.jupiter.api.Test
```
zaleslaw, line 4: agree, that test with zeroes is not the best solution here and in many same cases

api/src/test/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/LayerTest.kt
```diff
@@ -0,0 +1,90 @@
...
+            1 -> tf.constant((input as Array<Float>).toFloatArray())
+            2 -> tf.constant(input as Array<FloatArray>)
+            3 -> tf.constant(input as Array<Array<FloatArray>>)
+            4 -> tf.constant(input as Array<Array<Array<FloatArray>>>)
+            5 -> tf.constant(input as Array<Array<Array<Array<FloatArray>>>>)
+            else -> throw IllegalArgumentException()
+        }
+    }
+
+    protected fun runLayerInEagerMode(
```
zaleslaw, line 51: very useful function, we need something same to test layers in Eager Mode and in Graph mode too, great

zaleslaw (commented):
It's a great PR, but you need to wait to merge all minor tickets with different pool layers; this PR will last in this queue of PRs this week (I hope).

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/pooling/GlobalAvgPool1D.kt
mkaze: Well, the input shape is already included in the "NOTE". And probably I have removed the output shape just due to consistency with documentation of other pooling layers. I'll revert back this change if you want.

zaleslaw: Yes, please remove this documentation, I'm going to add this information for all layers, think it will be useful for the user.

api/src/test/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/LayerTest.kt
```diff
@@ -0,0 +1,90 @@
...
+    protected fun createFloatArray4D(shape: LongArray): Array<Array<Array<FloatArray>>> {
+        return Array(shape[0].toInt()) {
+            createFloatArray3D(shape.drop(1).toLongArray())
+        }
+    }
+
+    protected fun createFloatArray5D(shape: LongArray): Array<Array<Array<Array<FloatArray>>>> {
+        return Array(shape[0].toInt()) {
+            createFloatArray4D(shape.drop(1).toLongArray())
+        }
```
avan1235, line 37: Functions for creating array are already present in ShapeFunctions file @mkaze  Maybe try to use them in tests as they look they would be suitable

api/src/test/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/LayerTest.kt
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    private fun getInputOp(tf: Ops, input: Array<*>, inputDim: Int): Operand<Float> {
```
avan1235, line 40: And this @mkaze can be achieved I think with the helper functions present in ShapeFunctions which are used for casting arrays. Are they suitable for these tests?

avan1235 (changes requested):
Let's see at the comments as it seems for me like a code duplication as similar (even identical in result) functions are present in more general ShapeFunctions file

mkaze:
> Let's see at the comments as it seems for me like a code duplication as similar (even identical in result) functions are present in more general ShapeFunctions file

@avan1235 Yeah, you are correct; I was not aware of the similar changes you had committed a few days ago (actually, it seems you have refactored some of the codes into helper and extension functions in the past two weeks :smile: ). Actually, after I became aware of this I wanted to mention that there is this duplication and we should address it. I saw some common patterns in testing the layers which each of us is repeating to some extent in our implementations and my main goal is to refactor them so that we could use them in all of the related tests and avoid duplication. Some of them are (which may currently exist or not):

- Helper functions for creating N-D Arrays.
- Helper functions for casting Arrays to specific dimensionality.
- A single function to run a forward pass on a given instance of a layer and given an input data array of arbitrary dimensionality and return the output of the layer. (either in eager/graph mode)
- Verify the output of layer with the expected result.
- Verify the output of `computeOutputShape` method of layer with the expected output shape.
- Test config loader/saver functions of the layer.

I think most of these are already implemented; however, I think there are currently some duplication in the tests which could be further refactored and simplified, so that all layers' test use a common API regardless of the type of the layer (i.e. it does not matter if it's a pooling or convolution layer). @avan1235 I'll wait for your Conv3D PR to be merged and after that I'll review the tests to find out the opportunities for refactoring and see how this PR could be improved.

api/src/test/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/LayerTest.kt
zaleslaw: Agree, that there a lot of the functionality which doing the same or close things related to shapes/multidimensional arrays

zaleslaw:
@mkaze It's great suggestions about refactoring current testing functionality. As you see, when you joined the project there is no good solution for testing layers, especially layers with variables (via Graph mode). It was tested on high-level by examples (as integration tests) and that's all.

I'll create a separate ticket for this suggestion before we are moving forward with many new layers and architectures.

zaleslaw:
Really strange, but I have a compilation error\

```
:api:compileTestKotlin
11:21:24     e: /mnt/agent/work//api/src/test/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/GlobalAvgPooling2DTest.kt: (11, 9): Unresolved reference: assertGlobalAvgPool2DEquals
11:21:24     e: /mnt/agent/work//api/src/test/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/GlobalAvgPooling3DTest.kt: (11, 9): Unresolved reference: assertGlobalAvgPool3DEquals
```
Maybe some import are missed

mkaze:
@zaleslaw Sorry, I had to be clear there: I am currently working on it to refactor the tests and remove duplicated functionalities. I just merged it with master branch to update it with the latest changes. So it's not finished yet.
I think I'll use this PR for refactoring the testing for all layers and not just the global pooling layers, if you are agreed.

zaleslaw:
Yes, but maybe you will set up the WIP status for this PR 
If you want, you could work on all layers test improvements, but keep in mind that new layers could be added in the nearest future and will make this PR greater and greater, maybe better to finish it on the pooling tests field.

api/src/test/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/LayerTest.kt
```diff
@@ -0,0 +1,142 @@
...
+    private fun getInputOp(tf: Ops, input: Array<*>): Operand<Float> {
+        return when (input.shape.numDimensions()) {
+            1 -> tf.constant(input.map { it as Float }.toFloatArray())
+            2 -> tf.constant(input.cast2D<FloatArray>())
+            3 -> tf.constant(input.cast3D<FloatArray>())
+            4 -> tf.constant(input.cast4D<FloatArray>())
+            5 -> tf.constant(input.cast5D<FloatArray>())
+            else -> throw IllegalArgumentException("Inputs with more than 5 dimensions are not supported yet!")
+        }
+    }
```
mkaze, line 26: I am not sure if this could be further refactored or improved.

api/src/test/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/LayerTest.kt
```diff
@@ -0,0 +1,142 @@
...
+                     * Must specify at least one target to fetch or execute.
+                     * ```
+                     * However, using `layer.weights.isNotEmpty()` for the condition would not work
+                     * because according to the current implementation of `extractWeights` the layer
+                     * should belong to an actual model and therefore it throws an error. Hence, we
+                     * are using `paramCount` instead to determine whether the layer has any trainable
+                     * weights.
+                     */
+                    if (layer.paramCount > 0)
+                        kGraph.initializeGraphVariables(session)
```
mkaze, line 77: Why does `initializeGraphVariables` always expect some variables to be created and present? Couldn't it be modified so that it does not run the session if there is no variable to initialize? In that case, we can get rid of this `if` statement.

api/src/test/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/LayerTest.kt
```diff
@@ -0,0 +1,142 @@
...
+        val outputArray = getFloatArrayOfShape(outputShape).let {
+            when (outputShape.numDimensions()) {
+                1 -> it as Array<Float>
+                2 -> it.cast2D<FloatArray>()
+                3 -> it.cast3D<FloatArray>()
+                4 -> it.cast4D<FloatArray>()
+                5 -> it.cast5D<FloatArray>()
+                else -> throw IllegalArgumentException("Arrays with more than 5 dimensions are not supported yet!")
+            }
+        }
```
mkaze, line 114: Again, this is the best I could get to. Maybe it could be further improved.

api/src/test/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/PReLUTest.kt
```diff
@@ -5,111 +5,48 @@
...
-                    layer.build(tf, kGraph, inputShape)
-
-                    val inputOp = tf.constant(input)
-                    val isTraining = tf.constant(false)
-                    val numberOfLosses = tf.constant(1.0f)
-                    val output = layer.forward(tf, inputOp, isTraining, numberOfLosses).asOutput()
-                    kGraph.initializeGraphVariables(session)
-                    val outputTensor = session.runner().fetch(output).run().first()
+        // NOTE: It seems testing layers having trainable weights in Eager mode is not working.
+        //assertLayerOutputIsCorrect(layer, input, expected)
```
mkaze, line 35: @zaleslaw I remember you once mentioned that the Eager mode implementation in low-level API is buggy. Just to reconfirm that's actually the case? Because it really throws an error...

api/src/test/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/LayerTest.kt
```diff
@@ -0,0 +1,142 @@
...
+    protected fun assertLayerOutputIsCorrect(
+        layer: Layer,
+        input: Array<*>,
+        expectedOutput: Array<*>,
+        runMode: RunMode = RunMode.EAGER,
+    ) {
+        val output = when (runMode) {
+            RunMode.EAGER -> runLayerInEagerMode(layer, input)
+            RunMode.GRAPH -> runLayerInGraphMode(layer, input)
+        }
```
mkaze, line 103: The `output` would be of type `Tensor<*>`, but I am not sure if I should call its `close()` method at the end of this function to release the allocated resources. This is what the `Tensor` documentation says:

> WARNING: Resources consumed by the Tensor object must be explicitly freed by invoking the close() method when the object is no longer needed. For example, using a try-with-resources block:
> ```
> try (Tensor t = Tensor.create(...)) {
>     doSomethingWith(t);
> }
> ```

And this is what the documentation of `close()` method says:

> WARNING:This must be invoked for all tensors that were not been produced by an eager operation or memory will be leaked.
> The Tensor object is no longer usable after close returns.

mkaze (commented):
@zaleslaw @avan1235 I revised my changes and this is the result. With this, for all the layers,  we could use a single function call to check if their implementation (i.e. forward pass) is correct or not. Note that I have kept this PR intentionally small just to demonstrate the usefulness of refactoring; besides the GlobalAvgPool layers, I also used it in a layer with trainable weights (i.e. PReLU in this case) to show its applicability. But potentially this could be used for all the layers and to remove the duplicated/redundant functionalities.

Please take a look and share your opinions/comments. I have also left some self-review comments and it would be great if you could also have a look at them as well. Thanks!

mkaze:
Hey @zaleslaw, since I didn't hear from you I was wondering if you had time to review the changes I made for this PR. Please let me know if you have reviewed this, or if this PR should be postponed for a later time. Thanks!

zaleslaw:
It's in wip status, this is a reason why I not started the review, or do you need early feedback on that? Please clarify the status of the pr

mkaze:
@zaleslaw Yes, I'd like your feedback please. To me, it's finished unless you have other comments; sorry, maybe I should have been more clear about that (i.e. either explicitly requested a review, or removed the WIP status).

api/src/test/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/LayerTest.kt
zaleslaw: It's a good idea, please create an enhancement about it, I will handle your idea and answer there

api/src/test/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/LayerTest.kt
zaleslaw: better to do all closing of tensors manually, also the section `.use` could be used as an analogue of java try-with-resources

api/src/test/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/PReLUTest.kt
zaleslaw: eager mode in 1.15 runtime was limited and for complex things as layers I suggest to use GraphMode

api/src/test/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/PReLUTest.kt
```diff
@@ -124,65 +61,36 @@ class PReLUTest : ActivationLayerTest() {
...
+        EagerSession.create().use {
+            val tf = Ops.create()
+            val alphaTensor = alphaInitializer.initialize(
+                inputShape[2].toInt(),
+                inputShape[2].toInt(),
+                tf,
+                shapeOperand(tf, Shape.make(1, inputShape[2])),
+                "temp_init"
+            ).asOutput().tensor()
+            alphaTensor.copyTo(alpha)
```
zaleslaw, line 79: tensor should be closed manually

zaleslaw (changes requested):
Be careful with closing tensors, there are two way: to do it manually via `.close` or via` .use { ... } `(analogue of try-with-resources with the call of the `close `method under the hood for tensor if `tensor.use {...} `was written.

zaleslaw:
@mkaze I suppose this is not the final version of assert on layers, but it's good enough for the first iteration because we really need this to avoid a ton of duplicated code.

We have an idea related to this topic from @dosier in issue #145

zaleslaw:
@mkaze, please notify me when the PR will be ready for the final review and running TC

api/src/test/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/LayerTest.kt
zaleslaw: also, you should close it for the graph session, of course, and test the closing for eager mode (maybe we could avoid closing for it), maybe you will got an error about the tensor was closed twice or something related incorrect tensor handling

mkaze:
@zaleslaw Thanks! I made some additional changes according to your comments, which you can see here: 0ee4ed7. I think it's ready for your final review.

PR #105, opened 2021-06-07
Add missing case for Softmax layer to ModelLoader

mkaze:
As the title says.


zaleslaw:
Many thanks, I found a few missed layers here, just need to refactor these bloody two methods produced via copy-paste to avoid such issues in the future

Issue #79, opened 2021-05-31
[Video] Add Conv3D layer


zaleslaw:
We are missing some layers to support the export of models from Keras fully. One of them is the Conv3D layer. A detailed description of the layer can be found [here](https://keras.io/api/layers/convolution_layers/convolution3d/).

- [ ] add layer class to the  [package](https://github.com/JetBrains/KotlinDL/tree/master/api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/convolutional) using Conv2D as a [reference implementation](https://github.com/JetBrains/KotlinDL/blob/master/api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/convolutional/Conv2D.kt)
- [ ] add documentation
- [ ] add test
- [ ] add export function (see [ModelSaver.kt](https://github.com/JetBrains/KotlinDL/blob/master/api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/inference/keras/ModelSaver.kt)) 
- [ ] add import function (see [ModelLoader.kt](https://github.com/JetBrains/KotlinDL/blob/master/api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/inference/keras/ModelLoader.kt))
- [ ] (optional) create an example

This issue depends on the results of the #59

avan1235:
I can take it after the merge of #59

PR #114, opened 2021-06-14
Fix naming inconsistencies

### LINKED ISSUE #94: "Pool" or "Pooling" ?!
Currently, we are using a mix of both "Pool" and "Pooling" words in naming the variables, functions, classes and the files, as well as in constant string values. I think this might bring confusion and inconsistency. It would be a good idea to choose only one of these and use it all over the codebase. Of course, we can choose the other one only as required (e.g. for compatibility with Keras when loading models).

For comparison, this was also used to be the case in Keras/TF (and indeed [it creates confusion](https://stackoverflow.com/q/63006575/2099607) as some users thought "pool" and "pooling" are different!); however, to mitigate it to some degree TF developers now use "pool", and the "pooling" would be considered as an alias of "pool" (even in documentation links); although, with the exception of "average pooling" which still uses "pooling" (probably because they have used "average" instead of "avg" for clarity and therefore in this case decided to chose the longer version, i.e. "pooling" instead of "pool", for consistency).

Personally, I suggest to always use "MaxPool" and "AvgPool" in naming everything, from variables to files to values of string constants. And only use "Pooling" as necessary (e.g. for compatibility reasons).

### LINKED ISSUE #106: Some inconsistencies in naming and data types
The inconsistencies are as follows:

**ModelLoader.kt**

- The name of some of the layer creator functions follows the pattern `create[layer_name]Layer` while others don't use that "Layer" at the end. [suggestion for fixining this: refactor and always use "Layer" at the end for these functions.]
- The return type of some of the layer creator functions has been set as `Layer` while others use the type of the specific layer (e.g. Dense, or Flatten layer). [suggestion for fixing this: always use `Layer` as the return type of these functions.]

**ModelSaver.kt**

- Same as the first point for "ModelLoader.kt".

**Numerical types**

- We are using both `Long` and `Int` in an interchangeable way. For example, `AvgPool2D` and `MaxPool2D` are using the type `IntArray` for their public properties like `poolSize` (which is then converted to `Long` in `forward` method), while most of the other layers stick to `Long`. [suggestion for fixing this: not sure... on one hand it seems the underlying Java API expects `Long` most of the time, but on the other hand using `Long` in places where `Int` already works seems to be a bit redundant.]

mkaze:
Resolves #94, Resolves #106. This PR fixes naming inconsistencies as well as some formatting issues. Specifically, following items have been done:

 - [x] Use "Layer" word at the end of name of each layer creator function in both `ModelLoader.kt` and `ModelSaver.kt`
 - [x] Use "Keras" word in name of each layer creator function in `ModelSaver.kt`
 - [x] Use "Pool" instead of "Pooling" all over the code (except for: comments, the directory name "pooling" and value of constants in `KerasConstants.kt`)
 - [x] Reformat and reorganize part of the codes in `ModelLoader.kt`, `ModelSaver.kt` and `KerasConstants.kt` to keep relevant/similar sections close to each other.


zaleslaw:
@mkaze Thanks for making the code of export/import better and more clear! 
So, we will wait before #79 will be merged to apply these changes

zaleslaw:
Hi @mkaze I suggest to start from this PR and continue with the #103 as the second task
Now Conv3d is merged and you could updated your PR

mkaze:
@zaleslaw All right, I just updated this PR with recent changes on master branch. It's ready to be reviewed.

Issue #78, opened 2021-05-31
Add missed loader for the ReLU and ELU activation layers


zaleslaw:
Each layer should have an implementation for export/import from/to JSON configuration (Keras compatible format) (see [ModelLoader.kt ](https://github.com/JetBrains/KotlinDL/blob/master/api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/inference/keras/ModelLoader.kt) and [ModelSaver.kt](https://github.com/JetBrains/KotlinDL/blob/master/api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/inference/keras/ModelSaver.kt))

The saving functions for the ReLU  and ELU activation layer are missed

As an integration example, add a convolutional neural network to the examples package (in CNN package for example), it could be based on [improved LeNet model](https://github.com/JetBrains/KotlinDL/blob/master/examples/src/main/kotlin/examples/cnn/mnist/ModernLeNet.kt), train it on "mnist" or "fashion mnist" and add it as an integration test to the "examples" module tests

- [ ] add saving function for ReLU activation layer
- [ ] add saving function for ELU activation layer
- [ ] add an example
- [ ] convert this example to the integration test
- [ ] train network, tune it to achieve accuracy more than >70%

avan1235:
I can take if there are no volunteers for this issue and I was adding the ELU layer as I don't see other work for now for me and I'd like to try to export the trained model in kotlinDL to see how it works :wink:

zaleslaw:
I suppose you are enough mature for second-level tasks, let's leave it for newcomers, it's easy to add, but need to understand some important things.
If it will be not closed till the end of July, I'll notify you or fix it be myself

dosier:
I can volunteer for this issue. 
I already see there is support for loading **ELU** and **ReLU** layers in [ModelLoader.kt](https://github.com/JetBrains/KotlinDL/blob/master/api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/inference/keras/ModelLoader.kt)
So only the saving part needs to be done right?

dosier:
Nevermind my commits, we happened to work on it at the same time :D

mkaze:
[...]

PR #115, opened 2021-06-14
Add loaders for ELU, RELU activation layers (#78)

kokorins:
- missing activation layers loaders added
- two distinct examples with save/load added to examples folder (trying to reach 0.7 accuracy)

### DIFF
--- a/examples/src/main/kotlin/examples/inference/saveload/SaveLoadElu.kt
+++ b/examples/src/main/kotlin/examples/inference/saveload/SaveLoadElu.kt
@@ -1,0 +1,112 @@
+package examples.inference.saveload
+
+import examples.inference.lenet5
+import org.jetbrains.kotlinx.dl.api.core.Sequential
+import org.jetbrains.kotlinx.dl.api.core.activation.Activations
+import org.jetbrains.kotlinx.dl.api.core.initializer.GlorotNormal
+import org.jetbrains.kotlinx.dl.api.core.initializer.GlorotUniform
+import org.jetbrains.kotlinx.dl.api.core.layer.convolutional.Conv2D
+import org.jetbrains.kotlinx.dl.api.core.layer.convolutional.ConvPadding
+import org.jetbrains.kotlinx.dl.api.core.layer.core.Dense
+import org.jetbrains.kotlinx.dl.api.core.layer.core.Input
+import org.jetbrains.kotlinx.dl.api.core.layer.pooling.MaxPool2D
+import org.jetbrains.kotlinx.dl.api.core.layer.reshaping.Flatten
+import org.jetbrains.kotlinx.dl.api.core.loss.Losses
+import org.jetbrains.kotlinx.dl.api.core.metric.Metrics
+import org.jetbrains.kotlinx.dl.api.core.optimizer.SGD
+import org.jetbrains.kotlinx.dl.dataset.handler.NUMBER_OF_CLASSES
+import org.jetbrains.kotlinx.dl.dataset.mnist
+import java.io.File
+
+
+private const val MODEL_SAVE_PATH = "savedmodels/elu_lenet_saveload"
+
+private const val NUM_CHANNELS = 1L
+private const val IMAGE_SIZE = 28L
+private const val SEED = 12L
+
+private val kernelInitializer = GlorotNormal(SEED)
+private val biasInitializer = GlorotUniform(SEED)
+
+/**
+ * See [lenet5]. This just has Relu replaced for ELU on earlier layers for save/load test.
+ */
+private fun modifiedLenet5(): Sequential = Sequential.of(
+    Input(
+        IMAGE_SIZE,
+        IMAGE_SIZE,
+        NUM_CHANNELS,
+        name = "input_0"
+    ),
+    Conv2D(
+        filters = 32,
+        kernelSize = longArrayOf(5, 5),
+        strides = longArrayOf(1, 1, 1, 1),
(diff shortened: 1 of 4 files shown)


examples/src/main/kotlin/examples/inference/saveload/SaveLoadExample.kt
```diff
@@ -0,0 +1,46 @@
...
+import org.jetbrains.kotlinx.dl.api.core.callback.Callback
+import org.jetbrains.kotlinx.dl.api.core.history.BatchTrainingEvent
+import org.jetbrains.kotlinx.dl.api.core.history.TrainingHistory
+import org.jetbrains.kotlinx.dl.api.core.loss.Losses
+import org.jetbrains.kotlinx.dl.api.core.metric.Metrics
+import org.jetbrains.kotlinx.dl.api.core.optimizer.SGD
+import org.jetbrains.kotlinx.dl.dataset.Dataset
+import java.io.File
+
+object SaveLoadExample {
```
zaleslaw, line 15: Could you please add some KDocs here with the high-level description of this object (is it a helper object?)

examples/src/main/kotlin/examples/inference/LeNetModel.kt
```diff
@@ -27,7 +27,7 @@ private val biasInitializer = GlorotUniform(SEED)
 /**
  * Returns classic LeNet-5 model with minor improvements (Sigmoid activation -> ReLU activation, AvgPool layer -> MaxPool layer).
  */
-fun lenet5(): Sequential = Sequential.of(
+fun lenet5(sigmoidActivations:Activations = Activations.Relu): Sequential = Sequential.of(
```
zaleslaw, line 30: Is it a helpful change? Are you going to run it with another Activation function somewhere?

zaleslaw (commented):
Really great and interesting examples

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/inference/keras/ModelLoader.kt
```diff
@@ -398,7 +398,7 @@ private fun createActivationLayer(config: LayerConfig, name: String): Layer {
 
 private fun createReLULayer(config: LayerConfig, name: String): Layer {
     return ReLU(
-        maxValue = config.max_value!!.toFloat(),
+        maxValue = config.max_value?.toFloat(),
```
avan1235, line 401: shouldn't it be left as it was as here the config is supposed to have this value

examples/src/main/kotlin/examples/inference/LeNetModel.kt
avan1235: btw it's name is missleading, as it is activation, not sigmoid

examples/src/main/kotlin/examples/inference/saveload/SaveLoadExample.kt
avan1235: @kokorins SaveLoadExample should be imo a class as we see it can be parametrized

examples/src/main/kotlin/examples/inference/saveload/SaveLoadElu.kt
```diff
@@ -0,0 +1,30 @@
...
+import org.jetbrains.kotlinx.dl.api.core.Sequential
+import org.jetbrains.kotlinx.dl.api.core.activation.Activations
+import org.jetbrains.kotlinx.dl.api.core.loss.Losses
+import org.jetbrains.kotlinx.dl.api.core.metric.Metrics
+import org.jetbrains.kotlinx.dl.api.core.optimizer.SGD
+import org.jetbrains.kotlinx.dl.dataset.mnist
+import java.io.File
+
+
+private const val PATH_TO_MODEL = "savedmodels/elu_lenet_saveload"
```
avan1235, line 13: imo it's MODEL_SAVE_PATH as at the begging this file does not exist and it's going to be created after the save process (and then read, but saving is as important as reading)

avan1235 (commented):
Just my thoughts, think about them @kokorins. Btw @zaleslaw we can see that linting specificatin for code style is needed as e.g. there are no newlines at the end of files (while present in most of the files) and other inconsistencies in this area are present (while sould be pointed by some automatic rules)

examples/src/main/kotlin/examples/inference/LeNetModel.kt
kokorins: Agreed, just followed the comments for `Sigmoid activation -> ReLU activation` ;-)

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/inference/keras/ModelLoader.kt
kokorins: I'm not fully understand notation, maxValue is nullable, whereas other fields are not.

avan1235: I was thinking that it shouldn't be nullable in config (and then the exception would be expected on !! as it was) but now I see that it might be configured as null and it has a special meaning thought. Btw @zaleslaw  is an author of ReLU so he would be the best  source of answer here how should be a nullable value interpreted from config

kokorins:
Cant re-open this one, please continue with: https://github.com/JetBrains/KotlinDL/pull/128

PR #127, opened 2021-06-15
Add AbstractConv documentation

### LINKED ISSUE #118: Missed documentation in Conv classes hierarchy
After merging I recognized that the KDocs for the Conv class hierarchy are missed

Please add KDocs for all classes, methods, and fields in the following classes

- [ ] AbstractConv (class, properties, methods)
- [ ] Conv1D (methods)
- [ ] Conv2D (methods)
- [ ] Conv3D (methods)
- [ ] DepthwiseConv2D (methods)

I suppose that enough to define good KDoc for the parent (protected) method of property, the documentation is inherited by child methods

avan1235:
Resolves #118. Adding the documentation only for parent class seems to be good enough and that's the reason I did'n provide the documentation for `build`, `forward` etc. as the documentation is present in `Layer`.

@zaleslaw If you can try to look at the description of `AbstractConv` because I'm not sure if I was able to express my intention in a readable way so it maybe should be improved

### DIFF
--- a/api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/convolutional/AbstractConv.kt
+++ b/api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/convolutional/AbstractConv.kt
@@ -137,7 +137,6 @@ public abstract class AbstractConv(
      *  Defaults to the number of filter in convolutional layer. */
     protected open fun getOutputDepth(numberOfChannels: Long): Long = filtersInternal
 
-
     /**
      * Define the [kernelShape] by default from its [kernelSizeInternal],
      * [filtersInternal] and the given [numberOfChannels] from input Tensor.
@@ -155,7 +154,6 @@ public abstract class AbstractConv(
      */
     protected open fun computeBiasShape(numberOfChannels: Long): Shape =
         Shape.make(filtersInternal)
-
 
     /** Given a layer name specify its kernel name. */
     protected abstract fun kernelVarName(name: String): String
--- a/api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/convolutional/AbstractConv.kt
+++ b/api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/convolutional/AbstractConv.kt
@@ -19,6 +19,35 @@ import org.tensorflow.op.core.Variable
 import java.lang.IllegalArgumentException
 import kotlin.math.roundToInt
 
+/**
+ * Abstract Convolutional layer is a base block for building base types of convolutional layers
+ * of any dimensionality. It should simplify the internal calculations needed in most of
+ * the convolutional layers and abstract the process of naming weights for these layers. It keeps
+ * the actual implementation of convolutional layers e.i. the kernel and bias learnable variables
+ * that should be used in child classes in actual implementations of these layers. If the child class
+ * uses some values for its implementation in other form than it is kept in this child class,
+ * then this abstract class `internal` properties should keep the implementation values
+ * while the child class properties should keep the printable values that are more representative.
+ * But in most cases the `internal` and child values will be the same.
+ *
+ * @property filtersInternal number used by default in calculation of layer weights and i/o shapes
+ * @property kernelSizeInternal numbers used by default in calculation of layer weights and i/o shapes
+ * @property stridesInternal numbers used by default in calculation of layer weights and i/o shapes
+ * @property dilationsInternal numbers to keep for the dilations for implementation
+ * @property activationInternal activation used in [forward] operation implementation
+ * @property kernelInitializerInternal kernelInitializer used in actual kernel variable filling implementation
+ * @property biasInitializerInternal biasInitializer used in actual bias variable filling implementation
+ * @property kernelRegularizerInternal kernelRegularizer used in actual kernel variable filling implementation
+ * @property biasRegularizerInternal biasRegularizer used in actual bias variable filling implementation
+ * @property activityRegularizerInternal regularizer function applied to the output of the layer
+ * @property paddingInternal numbers to keep for the padding for implementation
+ * @property useBiasInternal flag if bias should be used during actual [forward] implementation
+ * @property kernelVariableName name of kernel used when no layer name is defined
+ * @property biasVariableName name of bias used when no layer name is defined
+ * @constructor Creates [AbstractConv] object
+ *
+ * @param name of the layer to name its variables
+ */
 public abstract class AbstractConv(
     protected val filtersInternal: Long,
     protected val kernelSizeInternal: LongArray,
@@ -36,13 +65,18 @@ public abstract class AbstractConv(
     protected val biasVariableName: String,
     name: String
 ) : Layer(name) {
-    // weight tensors
+
+    /** Tensor with learnable variables for kernel defined by internal shapes */
     protected lateinit var kernel: Variable<Float>
+
+    /** Tensor with learnable variables for bias defined by internal shapes */
     protected var bias: Variable<Float>? = null
 
-    // weight tensor shapes
-    protected lateinit var kernelShape: Shape
+    /** Shape of internal implementation of kernel variable */
     protected lateinit var biasShape: Shape
+
+    /** Shape of internal implementation of bias variable */
+    protected lateinit var kernelShape: Shape
 
     override fun build(tf: Ops, kGraph: KGraph, inputShape: Shape) {
         // Amount of channels should be the last value in the inputShape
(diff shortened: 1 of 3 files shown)


api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/convolutional/AbstractConv.kt
```diff
@@ -19,6 +19,35 @@ import org.tensorflow.op.core.Variable
 import java.lang.IllegalArgumentException
 import kotlin.math.roundToInt
 
+/**
+ * Abstract Convolutional layer is a base block for building base types of convolutional layers
+ * of any dimensionality. It should simplify the internal calculations needed in most of
+ * the convolutional layers and abstract the process of naming weights for these layers. It keeps
+ * the actual implementation of convolutional layers e.i. the kernel and bias learnable variables
```
dosier, line 26: minor typo in line 26: "e.i." should be " i.e."

avan1235: thanks for pointing this, just fixed. What about the essence of this doc - was it readable for you @dosier?

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/convolutional/AbstractConv.kt
```diff
@@ -19,6 +19,35 @@ import org.tensorflow.op.core.Variable
 import java.lang.IllegalArgumentException
 import kotlin.math.roundToInt
 
+/**
+ * Abstract Convolutional layer is a base block for building base types of convolutional layers
+ * of any dimensionality. It should simplify the internal calculations needed in most of
+ * the convolutional layers and abstract the process of naming weights for these layers. It keeps
+ * the actual implementation of convolutional layers i.e. the kernel and bias learnable variables
```
zaleslaw, line 26: , i.e.,

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/convolutional/AbstractConv.kt
```diff
@@ -19,6 +19,35 @@ import org.tensorflow.op.core.Variable
 import java.lang.IllegalArgumentException
 import kotlin.math.roundToInt
 
+/**
+ * Abstract Convolutional layer is a base block for building base types of convolutional layers
+ * of any dimensionality. It should simplify the internal calculations needed in most of
```
zaleslaw, line 24: It should simplify the internal calculations needed in most convolutional layers and abstract the naming weights for these layers.

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/convolutional/AbstractConv.kt
```diff
@@ -19,6 +19,35 @@ import org.tensorflow.op.core.Variable
...
+ * the convolutional layers and abstract the process of naming weights for these layers. It keeps
+ * the actual implementation of convolutional layers i.e. the kernel and bias learnable variables
+ * that should be used in child classes in actual implementations of these layers. If the child class
+ * uses some values for its implementation in other form than it is kept in this child class,
+ * then this abstract class `internal` properties should keep the implementation values
+ * while the child class properties should keep the printable values that are more representative.
+ * But in most cases the `internal` and child values will be the same.
```
zaleslaw, line 31: cases,

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/convolutional/AbstractConv.kt
```diff
@@ -19,6 +19,35 @@ import org.tensorflow.op.core.Variable
...
+ *
+ * @property filtersInternal number used by default in calculation of layer weights and i/o shapes
```
zaleslaw, line 33: add [ ] for property names

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/convolutional/AbstractConv.kt
```diff
@@ -19,6 +19,35 @@ import org.tensorflow.op.core.Variable
...
+ * @property kernelSizeInternal numbers used by default in calculation of layer weights and i/o shapes
+ * @property stridesInternal numbers used by default in calculation of layer weights and i/o shapes
+ * @property dilationsInternal numbers to keep for the dilations for implementation
+ * @property activationInternal activation used in [forward] operation implementation
+ * @property kernelInitializerInternal kernelInitializer used in actual kernel variable filling implementation
+ * @property biasInitializerInternal biasInitializer used in actual bias variable filling implementation
```
zaleslaw, line 39: strange to have twice kernelInitializerInternal kernelInitializer 
Better to have something like that "[kernelInitializer] An initializer for the convolution kernel."

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/convolutional/AbstractConv.kt
avan1235: done

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/convolutional/AbstractConv.kt
avan1235: done

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/convolutional/AbstractConv.kt
avan1235: done

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/convolutional/AbstractConv.kt
avan1235: done

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/convolutional/AbstractConv.kt
```diff
@@ -100,14 +132,50 @@ public abstract class AbstractConv(
...
+     *  Defaults to the number of filter in convolutional layer. */
+    protected open fun getOutputDepth(numberOfChannels: Long): Long = filtersInternal
+
+    /**
+     * Define the [kernelShape] and [biasShape] by default from its [kernelSizeInternal],
+     * [filtersInternal], [filtersInternal] and the given [numberOfChannels] from input Tensor.
+     *
+     * @param numberOfChannels for input of this layer
+     */
+    protected open fun computeMatricesShapes(numberOfChannels: Long) {
```
zaleslaw, line 145: I suggest splitting it into two small functions for each shape calculation, also rename it to something clear instead of computeMatricesShapes, more related to kernel and bias

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/convolutional/AbstractConv.kt
avan1235: done

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/convolutional/AbstractConv.kt
```diff
@@ -36,26 +65,29 @@ public abstract class AbstractConv(
     protected val biasVariableName: String,
     name: String
 ) : Layer(name) {
-    // weight tensors
+
+    /** Tensor with learnable variables for kernel defined by internal shapes */
     protected lateinit var kernel: Variable<Float>
+
+    /** Tensor with learnable variables for bias defined by internal shapes */
```
zaleslaw, line 72: Hmm, seems not clear with "learnable variables" We could see that this is a variable from the type. Maybe better to rename "Kernel weights" and "Bias weights".
All variable has shapes, it's redundant information

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/convolutional/Conv1D.kt
```diff
@@ -95,6 +95,8 @@ public class Conv1D(
         requireArraySize(dilations, 3, "dilations")
     }
 
+    /** Axis of height for which the extra dimension is added (unsqueezed) before actual
```
zaleslaw, line 98: Maybe better "Axis of height for which the extra dimension is added (unsqueezed) before actual convolution operation and the output from actual implementation are squeezed."

zaleslaw (changes requested):
@avan1235 Please look at my comments and rewrite some docs and probably refactor some functions

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/convolutional/AbstractConv.kt
avan1235: done

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/convolutional/AbstractConv.kt
avan1235: fixed

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/convolutional/Conv1D.kt
avan1235: thanks, fixed

avan1235:
@zaleslaw I hope I fixed everything, thanks for all suggestions (all applied). Btw please run the TCs as this adding docs PR now includes also the suggested functions refactoring

PR #139, opened 2021-06-18
Add RepeatVector layer

dosier:


dosier (commented):
This commit can be ignored. It is reverted later during a merge with master.

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/reshaping/RepeatVector.kt
```diff
@@ -0,0 +1,72 @@
...
+ * Input shape: `2D tensor of shape (num_samples, features)`.
+ *
+ * Output shape: `3D tensor of shape (num_samples, n, features)`.
+ *
+ * @property n Repetition factor.
+ * @property [name] Custom layer name.
+ * @constructor Creates [RepeatVector] object.
+ *
+ * @author Stan van der Bend
+ * @since 0.2
```
avan1235, line 28: I suppose you mean 0.3

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/reshaping/RepeatVector.kt
```diff
@@ -0,0 +1,72 @@
...
+ */
+public class RepeatVector(
+    public val n: Int,
+    name: String = ""
+) : Layer(name) {
+
+    override fun build(tf: Ops, kGraph: KGraph, inputShape: Shape) {
```
avan1235, line 35: I propose
```kotlin
override fun build(tf: Ops, kGraph: KGraph, inputShape: Shape) = Unit
```
as this approach describes itself and doesn't need any comments that brings nothing to reader (as Kotlin gives us the bility to write such code, not like in Java)

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/reshaping/RepeatVector.kt
```diff
@@ -0,0 +1,72 @@
...
+        //left empty
+    }
+
+    override fun computeOutputShape(inputShape: Shape): Shape {
+        require(inputShape.numDimensions() == 2) { "input tensor must have 2 dimensions" }
```
avan1235, line 40: I saw only the phrases starting from big letter as checks reports. Keep the convention to have a well documented library

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/reshaping/RepeatVector.kt
```diff
@@ -0,0 +1,72 @@
...
+        val tensorShape = TensorShape(inputShape)
+        // TODO: maybe make `n` of type Long?
```
avan1235, line 42: imo you should as for now most of the ocnfiguration properties seems to have Long type.
btw why not just 
```kotlin
return Shape.make(inputShape[0], n, inputShape[1])
```
the `tensorShape` is unnecessary.

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/reshaping/RepeatVector.kt
```diff
@@ -0,0 +1,72 @@
...
+        return Shape.make(tensorShape[0], n.toLong(), tensorShape[1])
+    }
+
+    override fun forward(
+        tf: Ops,
+        input: Operand<Float>,
+        isTraining: Operand<Boolean>,
+        numberOfLosses: Operand<Float>?
+    ): Operand<Float> {
+        return tf.repeat(input, n)
```
avan1235, line 52: there are no benefits from creating the `Ops.repeat` as separate function (it's not going to be reused anywhere). Move your implementation to only `forward` as it is short enough

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/reshaping/RepeatVector.kt
```diff
@@ -0,0 +1,72 @@
...
+        set(value) = assignWeights(value)
+
+    override val hasActivation: Boolean get() = false
+
+    override val paramCount: Int get() = 0
+
+    override fun toString(): String {
+        return "RepeatVector"
+    }
+}
```
avan1235, line 72: you can fix this missing newline with other changes

api/src/test/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/RepeatVectorLayerTest.kt
```diff
@@ -0,0 +1,55 @@
+/*
+ * Copyright 2020 JetBrains s.r.o. and Kotlin Deep Learning project contributors. All Rights Reserved.
```
avan1235, line 2: Once again 2021 😉

api/src/test/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/RepeatVectorLayerTest.kt
```diff
@@ -0,0 +1,55 @@
...
+import org.jetbrains.kotlinx.dl.api.core.KGraph
+import org.jetbrains.kotlinx.dl.api.core.layer.reshaping.RepeatVector
+import org.jetbrains.kotlinx.dl.api.core.shape.toIntArray
+import org.junit.jupiter.api.Assertions
+import org.junit.jupiter.api.Test
+import org.tensorflow.Graph
+import org.tensorflow.Output
+import org.tensorflow.op.Ops
+
+/**
```
avan1235, line 17: I see no comments in tests classes. Keep the convention imo to have clean code in project

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/reshaping/RepeatVector.kt
```diff
@@ -0,0 +1,72 @@
...
+ * @property [name] Custom layer name.
+ * @constructor Creates [RepeatVector] object.
+ *
+ * @author Stan van der Bend
+ * @since 0.2
+ */
+public class RepeatVector(
+    public val n: Int,
+    name: String = ""
+) : Layer(name) {
```
avan1235, line 33: you should check in init block if `n >= 1` like 
```kotlin
init {
    require(n >= 1) { "Number of repetitions (n) in RepeatVector should be positive but got $n" }
}
```

api/src/test/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/RepeatVectorLayerTest.kt
```diff
@@ -0,0 +1,55 @@
...
+ * A test for the [RepeatVector] layer.
+ *
+ * @author Stan van der Bend
+ */
+internal class RepeatVectorLayerTest {
+
+    @Test
+    fun `test output shape`(){
```
avan1235, line 25: file needs reformatting (auto format in Intellij with Ctrl+Alt+L)

api/src/test/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/RepeatVectorLayerTest.kt
```diff
@@ -0,0 +1,55 @@
...
+        val actual = y.tensor().copyTo(Array(3) { Array(layer.n) { FloatArray(3) } })
+        val expected = arrayOf(
+            arrayOf(floatArrayOf(0F, 1F, 2F), floatArrayOf(0F, 1F, 2F)),
+            arrayOf(floatArrayOf(0F, 1F, 2F), floatArrayOf(0F, 1F, 2F)),
+            arrayOf(floatArrayOf(0F, 1F, 2F), floatArrayOf(0F, 1F, 2F))
+        )
+        Assertions.assertArrayEquals(expected, actual)
+    }
+
+    private operator fun RepeatVector.invoke(input : Array<FloatArray>) : Output<Float> {
```
avan1235, line 46: Very good idea from my perspective but maybe open a separate issue for this or join [this discussion](https://github.com/JetBrains/KotlinDL/discussions/132) with your improvement to test readability and some proposal. Btw it should be done for the `Layer` and then reused in all tests in the way you did this.

avan1235: btw clean this code a little bit (I don't see the need to refer to `Shape` as `org.tensorflow.Shape`) and let's see that the block with single value `{ return value }` can be reduced to `= value` (especially when you use the `.let { }`)

avan1235 (changes requested):
I leave m thoughts, don't worry about them a lot but try to get from them as much as you can as in my opinion this code needs most of these fixes but by doing this you can learn creating cleaner code

avan1235:
> Commits [78196a8](https://github.com/JetBrains/KotlinDL/commit/78196a88018bad65df0adbf925ffc52862bd6abf) and [6347d65](https://github.com/JetBrains/KotlinDL/commit/6347d656d898fde3d2084e9b9489f6cddf1ed150) can be ignored. They are reverted in [208cd56](https://github.com/JetBrains/KotlinDL/commit/208cd56a913a2c85e6bafe96a70e51aef83d5d71)

Just FYI your commits are squashed before merge so I think the reviewer looks only on the final version of commit so the steps that you did are not so important - the final result counts only 😉

dosier:
@avan1235 thanks a lot for your comments. This is my first time contributing to an open source project and your comments were really thoughtful, will make changes and keep in mind for future contributions.

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/reshaping/RepeatVector.kt
dosier: I am a bit confused on this point. I wasn't sure on how to handle this assertion, so I used something I saw used elsewhere in the project. Any suggested change?

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/reshaping/RepeatVector.kt
dosier: @avan1235 Nice! I never thought of that. I copied that piece from another similar Layer class. We might wanna convert all of the cases to this. It's a lot cleaner.

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/reshaping/RepeatVector.kt
dosier: `Shape` doesn't have operator functions for get, I could do:
```Kotlin
        val input = inputShape.toLongArray()
        return Shape.make(input[0], n.toLong(), input[1])
```

mkaze: @dosier You can simply use `size`:
```kotlin
return Shape.make(inputShape.size(0), n.toLong(), inputShape.size(1))
```

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/reshaping/RepeatVector.kt
avan1235: I was only thinking about a small typo here so just write 
```kotlin
require(inputShape.numDimensions() == 2) { "Input tensor must have 2 dimensions but got ${inputShape.numDimensions()}" }
```
as I understand one of KotlinDL goals i sto provide readable errors for user that can be easily fixed

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/reshaping/RepeatVector.kt
dosier: @mkaze ah of course, thanks

zaleslaw:
Hi, @dosier is it ready for review?

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/inference/keras/ModelSaver.kt
```diff
@@ -530,6 +532,16 @@ private fun createKerasFlattenLayer(layer: Flatten): KerasLayer {
     return KerasLayer(class_name = LAYER_FLATTEN, config = configX)
 }
 
+private fun createKerasRepeatVectorLayer(layer: RepeatVector): KerasLayer {
+    val configX = LayerConfig(
```
zaleslaw, line 536: Don't forget trainable status

api/src/test/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/RepeatVectorLayerTest.kt
```diff
@@ -0,0 +1,50 @@
...
+import org.junit.jupiter.api.Test
+import org.tensorflow.Graph
+import org.tensorflow.Output
+import org.tensorflow.Shape
+import org.tensorflow.op.Ops
+
+internal class RepeatVectorLayerTest {
+
+    @Test
+    fun `test output shape`() {
```
zaleslaw, line 21: please make it more consistent, rename it without ``

api/src/test/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/RepeatVectorLayerTest.kt
```diff
@@ -0,0 +1,50 @@
...
+        val expected = arrayOf(
+            arrayOf(floatArrayOf(0F, 1F, 2F), floatArrayOf(0F, 1F, 2F)),
+            arrayOf(floatArrayOf(0F, 1F, 2F), floatArrayOf(0F, 1F, 2F)),
+            arrayOf(floatArrayOf(0F, 1F, 2F), floatArrayOf(0F, 1F, 2F))
+        )
+        Assertions.assertArrayEquals(expected, actual)
+    }
+
+    // TODO: generalise this for Layer, see https://github.com/JetBrains/KotlinDL/issues/145
+    private operator fun RepeatVector.invoke(input: Array<FloatArray>): Output<Float> = Ops.create().let { tf ->
```
zaleslaw, line 43: maybe you could make a util internal extension function for the Layer class?

zaleslaw (changes requested):
Please fix a few minor things

api/src/test/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/RepeatVectorLayerTest.kt
dosier: maybe I could, u mean as described in #145 ? 
What should I do for now though, remove the function?

zaleslaw: So at this moment, you could leave it as is, I suppose.
The improvement related to #145 should be implemented in the separate PR and discussed with other contributors faced with this problem.

PR #146, opened 2021-06-21
Add UpSampling layers

### LINKED ISSUE #143: Add UpSampling layers
Currently, the support for UpSampling layers is missing and it would be great if we could add support for these layers as well. The desired PR addressing this issue should include:

- [ ]  Implementation of layers classes named as `UpSampling1D`, `UpSampling2D` and `UpSampling3D`.
- [ ]  Common hierarchy of all UpSampling layers with abstract class with the common functionality
- [ ]  Documentation of layer and all non-private methods
- [ ]   JUnit tests in `api ` module
- [ ]  Support for export of layers to JSON (see [ModelSaver.kt][1])
- [ ]  Support for import of layers from JSON (see [ModelLoader.kt][2])

Also, if needed, you can take a look at Keras documentation for these layers: [UpSampling1D](https://keras.io/api/layers/reshaping_layers/up_sampling1d), [UpSamplling2D](https://keras.io/api/layers/reshaping_layers/up_sampling2d) and [UpSampling3D](https://keras.io/api/layers/reshaping_layers/up_sampling3d). 

**NOTE:** for the moment, there is no need to add support for "data format" (i.e. channels last vs. channels first) in your PR; you can simply assume the channels are always in the last dimension.

[1]: https://github.com/JetBrains/KotlinDL/blob/master/api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/inference/keras/ModelSaver.kt

[2]: https://github.com/JetBrains/KotlinDL/blob/master/api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/inference/keras/ModelLoader.kt

mkaze:
Resolves #143. This PR adds an abstract base class for UpSampling layers and implements UpSampling1D, UpSampling2D and UpSampling3D layers.

- [x] Implement abstract base class for UpSampling layers.
- [x] Implement `UpSampling1D` layer class.
- [x] Implement `UpSampling2D` layer class.
- [x] Implement `UpSampling3D` layer class.
- [x] Add support for exporting layers' config to JSON.
- [x] Add support for importing layers from their JSON config.
- [x] Add unit tests for all UpSampling layers.


mkaze:
@zaleslaw This PR is ready to be reviewed.

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/reshaping/AbstractUpSampling.kt
```diff
@@ -0,0 +1,66 @@
...
+    )
+    val repeated = splits.map { tf.tile(it, multiples) }
+    // The following check is due to the fact the `tf.concat` raise an error
+    // if only one tensor is given as its input.
+    if (repeated.size == 1)
+        return repeated[0]
+    return tf.concat(repeated, tf.constant(axis))
+}
+
+public enum class InterpolationMethod(internal val methodName: String) {
```
zaleslaw, line 62: Add KDocs

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/reshaping/AbstractUpSampling.kt
```diff
@@ -0,0 +1,66 @@
...
+
+package org.jetbrains.kotlinx.dl.api.core.layer.reshaping
+
+import org.jetbrains.kotlinx.dl.api.core.KGraph
+import org.jetbrains.kotlinx.dl.api.core.layer.Layer
+import org.tensorflow.Operand
+import org.tensorflow.Shape
+import org.tensorflow.op.Ops
+
+public abstract class AbstractUpSampling(
```
zaleslaw, line 14: Add KDocs for the class and for fields

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/reshaping/AbstractUpSampling.kt
```diff
@@ -0,0 +1,66 @@
...
+    ): Operand<Float> {
+        return upSample(tf, input)
+    }
+
+    protected abstract fun upSample(tf: Ops, input: Operand<Float>): Operand<Float>
+
+    protected abstract fun computeUpSampledShape(inputShape: Shape): Shape
+}
+
+public fun repeat(tf: Ops, value: Operand<Float>, repeats: Int, axis: Int): Operand<Float> {
```
zaleslaw, line 48: add KDocs

zaleslaw: Maybe it should be protected or internal

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/reshaping/UpSampling2D.kt
```diff
@@ -0,0 +1,73 @@
...
+    protected override fun upSample(tf: Ops, input: Operand<Float>): Operand<Float> {
+        val inputShape = input.asOutput().shape()
+        val newSize = tf.constant(
+            intArrayOf(
+                inputShape.size(1).toInt() * size[0],
+                inputShape.size(2).toInt() * size[1]
+            )
+        )
+        return when(interpolation) {
+            InterpolationMethod.NEAREST -> tf.image.resizeNearestNeighbor(input, newSize)
```
zaleslaw, line 64: Really strange that UpSampling layers have so different implementations and level of interpolation usage

zaleslaw (changes requested):
So, what implementation did you use as a reference? The internals of Keras seems different...

mkaze:
@zaleslaw Well, I used Keras implementation as reference. The implementation for 2D layer is the same (i.e. both Keras and my implementation use low-level image resizing methods). For the 1D and 3D layers, Keras defines a helper function for repeating the elements of a tensor along a specific axis and then use it to implement those layers. 

I also did the same (though, my implementation of the repeat function is a little bit different but the overall strategy is exactly the same) and used the repeat function in implementation of 3D layer; however, I didn't use it for 1D layer and instead opted to use the resize method of images for 1D layer as well (because basically you can view a 1D data as a 2D data as well). 

Personally, I don't think there is anything confusing about this, but anyways I can change the implementation of 1D layer to use the repeat function instead if that makes it more clear.

mkaze:
@zaleslaw The docs were improved and implementation of 1D layer was refactored to use the `repeats` helper function instead of image resizing approach.

mkaze:
@zaleslaw I decided to remove the redundant `computeUpSampledShape` method and directly use `computeOutputShape` instead.

zaleslaw:
Great PR and tests!

Issue #135, opened 2021-06-16
Change the TrainableModel::summary API to return `ModelSummary`


avan1235:
In current implementation the [TrainableModel::summary](https://github.com/JetBrains/KotlinDL/blob/2e529a89058617e22fd81e783c7b1af07c1cbfa9/api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/TrainableModel.kt#L317) return the list with the descriptions for the following layers.

This information is readable but doesn't allow for easy transformation of model summary gathered data as it is give as raw String. Additionally, the implementation classes log the provided layers description in more readable way, but it is only printed on stdio and not available at least as pretty formatted string that could bu printed later. The problem is more annoying in jupyter notebooks where we don't have an access to stdio and so the logged pretty summary is not visible.

I would propose creating some simple `ModelSummary` data class that would be responsible for holding the model summary data and would have some good implementation of `toString()` method that could be easily printed. It would allow for creating custom printing methods for `ModelSummary` as well as good keeping of model summary data as a single object instead of `List<String>`

zaleslaw:
Hi, @knok16. Give me a few days to think about it; I'll return to you in a couple of days.

zaleslaw:
Thanks for lighting the problem @avan1235. I totally agree about that the existing solution is not good for notebooks.
So great to have a few data classes like @knok16 proposed.

I suppose the proposed solution should contain the following things:

- [ ] data classes for Model and layer summary
- [ ] printSummary and logSummary functions (logging could print on a console too)
- [ ] tests for Sequential, Functional models and a few popular layers like Conv2D, ZeroPadding2D 
- [ ] refactoring for all examples which contain it.summary() to it.logSummary() for consistency.
- [ ] 1 executed notebook with the built artifact from the proposed PR to estimate how pretty it is

knok16:
Can I take it?

zaleslaw:
@knok16 Sure, it's yours

knok16:
changes for pp 1-3 https://github.com/JetBrains/KotlinDL/pull/216
will do p.4 in separate PR, to not clutter this one

zaleslaw:
Mostly closed in the #216 
I will release an artifact, load it to the Maven Central and make a double check on the Datalore and Jupyter notebooks

zaleslaw:
I've checked the 0.3.0-alpha-4 artifact, print summary works as expected and this issue has been fixed.

PR #147, opened 2021-06-21
Move tutorials to notebooks

### LINKED ISSUE #119: Convert tutorials in md format to Jupyter Notebooks with Kotlin kernel
Currently, we have tutorials in "/docs" folder with snippets of code, we need to move forward and convert them to the runnable notebooks (probably based on the existing examples stored in the release branch)

avan1235:
Closes #119

I prepared extra badges for each tutorial to load them automatically to [nbviewer](https://nbviewer.jupyter.org/) basing the configuration on the configs from lets-plot library examples.

We can run executable version of notebook just by single click on binder or browse it not in GitHub version by clicking on nbviewer. The preview should also be loaded on GitHub in the same way but loading process fails sometimes and manual reload is needed. Btw I linked refs to totorials to nbviewer location of files in markdown as this seems to be more readable and stable but they are still accessible from GitHub (FYI the nbviewer and binder use only files stored on repository so all files are kept only in repository).

For now, all refs are created to point to my fork, I will change it before merge to master to make it working for now (as there are no notebooks in master KotlinDL branch that the badges should point to). The change includes changing `avan1235` to `JetBrains`.

I encourage you to do the review by starting from the main [README.md on my fork](https://github.com/avan1235/KotlinDL/tree/notebooks), the nbviewer version of [quickstart guide](https://nbviewer.jupyter.org/github/avan1235/KotlinDL/blob/notebooks/docs/quick_start_guide.ipynb) or just from the [docs directory](https://github.com/avan1235/KotlinDL/tree/notebooks/docs) from my fork if you wish to use bare GitHub.

The changes includes some small changes in notebooks code to make it working and more readable in my opinion so feel free to comment with your opinion.


zaleslaw:
First of all, it's really great and more interactive way to have and run tutorials. Thanks for that implementation.

My user experience:

1. Binder initialization is slow (maybe we need to add in Readme. md) some docs about its architecture, running docker, loading software
2. Nbviewer integration is good and works
3. Github displays notebooks really bad
4. I could not read tutorials as plain text to copy some code in my idea on Github. I need to go nbviewer
5. Summary method in notebooks is ugly

It improves user experience, of course, let's think about contributors:
1. The content in notebooks is not equal to the code/text in the `/docs` folder. It's not easy to review
2. Contributing and fixing typos in markdown files is easier than in notebooks
3. What if we need different software (for example, different TF or Keras versions)? Could we set it differently for different notebooks?

I suggest splitting your improvements into a few parts for easier reviewing:
1. Fix some things in the code of tutorials (also, it could be adopted in examples too)
2. Propose changes in the text (probably I will need to ask for a review from our copy-editors, they do the same for the current version of tutorials)
3. Convert the content to notebooks (or copy content to the notebooks) with minor changes related to the notebooks (let's plot support, different summary method, and so on).

At this point, I think that we need to keep both, plain md tutorials (for IDEA users) and notebooks

avan1235:
> First of all, it's really great and more interactive way to have and run tutorials. Thanks for that implementation.
> 
> My user experience:
> 
>     1. Binder initialization is slow (maybe we need to add in Readme. md) some docs about its architecture, running docker, loading software

It's slow only on the first run of the environment as I have read [here](https://discourse.jupyter.org/t/how-to-reduce-mybinder-org-repository-startup-time/4956)

>     2. Nbviewer integration is good and works
> 
>     3. Github displays notebooks really bad

Yeah I saw this problem but hope that it is a matter of some time when it becomes more and more popular and GitHub fixes some bugs.

>     4. I could not read tutorials as plain text to copy some code in my idea on Github. I need to go nbviewer

They shouldn't be visible as plain text, I supposed GitHub will render them properly as notebooks every time and then from these notebooks we should be able to copy code/text.

>     5. Summary method in notebooks is ugly

That's why I created an issue #135 (one of them as there are more)
 
> It improves user experience, of course, let's think about contributors:
> 
>     1. The content in notebooks is not equal to the code/text in the `/docs` folder. It's not easy to review

I haven't thought about review time but it is possible to review the GitHub PR in IntelliJ and when working with notebooks we should use another plugin in IntelliJ for jupyter notebooks - then they are readable and editable like in nbviewer or datalore. 

>     2. Contributing and fixing typos in markdown files is easier than in notebooks

Yes, the plugin for IntelliJ would be required but it seems that then everybody could contribude as it looks then as Markdown files

>     3. What if we need different software (for example, different TF or Keras versions)? Could we set it differently for different notebooks?

Now it is not possible. We can find the official discussion [here](https://discourse.jupyter.org/t/allow-for-multiple-different-dependencies-per-repository/4109)
 
> I suggest splitting your improvements into a few parts for easier reviewing:
> 
>     1. Fix some things in the code of tutorials (also, it could be adopted in examples too)
> 
>     2. Propose changes in the text (probably I will need to ask for a review from our copy-editors, they do the same for the current version of tutorials)
> 
>     3. Convert the content to notebooks (or copy content to the notebooks) with minor changes related to the notebooks (let's plot support, different summary method, and so on).

May be a little bit annoying copy-paste work as Markdown files had some lines wrapped and will be hard to recoever/spot differences. I will try to work on it in this way but reversing the changes to Markdown now seems to me as unneeded work.

> At this point, I think that we need to keep both, plain md tutorials (for IDEA users) and notebooks

Let's try the IDEA plugin for jupyter notebooks.

P.S. I just realised that there is no plugin for Jupyter but rather an Ultimate version of IDEA. That's not the best approach btw so maybe some extra reasearch would be needed if we really would like to use notebooks?

zaleslaw:
Thanks for the good and clear comments on my questions, dear @avan1235 so, I need to think about it more and discuss it with colleagues, so after [vacation](https://github.com/JetBrains/KotlinDL/discussions/154) let's revisit this PR and define the next steps regarding the notebooks.

I will share with you later yet one approach of building tutorials from kotlinx.serialisation and multik projects based on tests/text preprocessing and markdown (but looks like your approach is better)

avan1235:
Ok, I hope to get to know some other approaches, maybe we will find the most suitable here. Btw I will try to create a separate PR with some changes for now for .md tutorial and examples as I think this PR can take longer time in such scenario 😉

PR #163, opened 2021-07-27
Bugfix/161 correct temporary dataset path

### LINKED ISSUE #161: Incorrect temporary folder for the cat-vs-dogs dataset
WE have incorrect paths in the [embeddedDatasets.kt](https://github.com/JetBrains/KotlinDL/blob/master/api/src/main/kotlin/org/jetbrains/kotlinx/dl/dataset/embeddedDatasets.kt#L276): the temporary data.zip is loaded to the `datasets/catdogs/data.zip` instead of `datasets/dogs-vs-cats/data.zip` 

1. Fix the issue
2. Refactor String constants moving common parts to one common constant
3. Check that the loading is correct with one of [these examples](https://github.com/JetBrains/KotlinDL/tree/master/examples/src/main/kotlin/examples/cnn/dogscats) in the `cache `directory

funkstermonster:
Now the data.zip is downloaded to the correct path.
This fixes the issue: Incorrect temporary folder for the cats-vs-dogs dataset [LINK](https://github.com/JetBrains/KotlinDL/issues/161)
fixes #161

### DIFF
--- a/api/src/main/kotlin/org/jetbrains/kotlinx/dl/dataset/embeddedDatasets.kt
+++ b/api/src/main/kotlin/org/jetbrains/kotlinx/dl/dataset/embeddedDatasets.kt
@@ -262,7 +262,7 @@ public fun cifar10Paths(cacheDirectory: File = File("cache")): Pair<String, Stri
 }
 
 /** Path to the Dogs-vs-Cats dataset. */
-private const val DOGS_CATS_IMAGES_ARCHIVE: String = "datasets/catdogs/data.zip"
+private const val DOGS_CATS_IMAGES_ARCHIVE: String = "datasets/dogs-vs-cats/data.zip"
 
 /** Returns path to images of the Dogs-vs-Cats dataset. */
 public fun dogsCatsDatasetPath(cacheDirectory: File = File("cache")): String =
--- a/api/src/main/kotlin/org/jetbrains/kotlinx/dl/dataset/embeddedDatasets.kt
+++ b/api/src/main/kotlin/org/jetbrains/kotlinx/dl/dataset/embeddedDatasets.kt
@@ -278,12 +278,15 @@ public fun dogsCatsDatasetPath(cacheDirectory: File = File("cache")): String =
 /** Path to the subset of Dogs-vs-Cats dataset. */
 private const val DOGS_CATS_SMALL_IMAGES_ARCHIVE: String = "datasets/small_catdogs/data.zip"
 
+/** Path to the Data.zip url to be downloaded from. */
+private const val DOGS_CATS_SMALL_IMAGES_DOWNLOAD_URL: String = "https://kotlindl.s3.amazonaws.com/datasets/small_catdogs/data.zip"
+
 /** Returns path to images of the subset of the Dogs-vs-Cats dataset. */
 public fun dogsCatsSmallDatasetPath(cacheDirectory: File = File("cache")): String =
     unzipDatasetPath(
         cacheDirectory,
-        loadFile(cacheDirectory, DOGS_CATS_SMALL_IMAGES_ARCHIVE),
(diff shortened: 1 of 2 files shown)


examples/src/test/kotlin/examples/cnn/dogscats/CatDogTestSuite.kt
```diff
@@ -5,10 +5,8 @@
 
 package examples.cnn.dogscats
 
-import org.junit.jupiter.api.Disabled
 import org.junit.jupiter.api.Test
 
-@Disabled
```
zaleslaw, line 11: No needs to remove Disabled annotation here. All these examples are too heavy and take a lot of time during usual checks.

api/src/main/kotlin/org/jetbrains/kotlinx/dl/dataset/embeddedDatasets.kt
```diff
@@ -262,13 +262,16 @@ public fun cifar10Paths(cacheDirectory: File = File("cache")): Pair<String, Stri
 }
 
 /** Path to the Dogs-vs-Cats dataset. */
-private const val DOGS_CATS_IMAGES_ARCHIVE: String = "datasets/catdogs/data.zip"
+private const val DOGS_CATS_IMAGES_ARCHIVE: String = "datasets/dogs-vs-cats/data.zip"
+
+/** Path to the Data.zip url to be downloaded from. */
+private const val DOGS_CATS_IMAGES_DOWNLOAD_URL: String = "https://kotlindl.s3.amazonaws.com/datasets/catdogs/data.zip"
```
zaleslaw, line 268: I recognized that we have the same problem with the paths for the small subset of this dataset.
Could you please do the same refactoring for the `DOGS_CATS_SMALL_IMAGES_ARCHIVE `constant and `dogsCatsSmallDatasetPath `method?

zaleslaw (changes requested):

api/src/main/kotlin/org/jetbrains/kotlinx/dl/dataset/embeddedDatasets.kt
funkstermonster: Done, also extended the extraction logic.

funkstermonster:
Hi @zaleslaw! Did you have a chance to take a look at it?

api/src/main/kotlin/org/jetbrains/kotlinx/dl/dataset/embeddedDatasets.kt
```diff
@@ -318,6 +324,12 @@ private fun unzipDatasetPath(cacheDirectory: File, archive: File, dirRelativePat
     if (!dataDirectory.exists()) {
         Files.createDirectories(dataDirectory.toPath())
 
+        extractFromZipArchiveToFolder(archive.toPath(), toFolder)
```
zaleslaw, line 327: it looks like copy-paste and from the "else" block and could be refactored to the one function

zaleslaw: if I understand correctly the proposed logic the same piece of code will be executed in if and else branches and could be moved below after if-else condition

zaleslaw: But I also need to add that logic was to download and unzip the archive only if the data folder does not exist. Suppose the data folder exists no need to download it again. Yes, we have no special validation here for data: is it the correct dataset or maybe not.

zaleslaw (changes requested):
Looks like the added functionality will download the dataset again and again. We tried to avoid it.

zaleslaw:
Fixed in #235

PR #183, opened 2021-08-09
Add TimeStopping callback

### LINKED ISSUE #174: Add TimeStopping callback
It's an exciting idea to limit the training via the maximum limit of training time.

The idea is implemented in [tf-addons](https://www.tensorflow.org/addons/api_docs/python/tfa/callbacks/TimeStopping) and could be adopted in KotlinDL

The desired PR addressing this issue should include:

- [ ]  Implementation of TimeStopping callback
- [ ]  Documentation 
- [ ]  Integration test with callback (see [LeNetWithEarlyStoppingCallback][1] )

[1]: https://github.com/JetBrains/KotlinDL/blob/master/examples/src/main/kotlin/examples/cnn/mnist/advanced/LeNetWithEarlyStoppingCallback.kt

P.S. If you want to take this ticket, please leave the comment below
P.P.S Read the [Contributing Guidelines](https://github.com/JetBrains/KotlinDL/blob/master/CONTRIBUTING.md).

The reference implementation could be taken from [tensorflow-addons](https://github.com/tensorflow/addons/blob/v0.13.0/tensorflow_addons/callbacks/time_stopping.py)

mkaze:
Resolves #174.

This PR adds `TimeStopping` callback. It includes an example, `LetNetWithTimeStoppingCallback.kt`, serving as the integration test.


examples/src/main/kotlin/examples/cnn/mnist/advanced/LeNetWithTimeStoppingCallback.kt
```diff
@@ -0,0 +1,129 @@
...
+fun lenetWithTimeStoppingCallback() {
+    val (train, test) = mnist()
+
+    lenet5Classic.use {
+        val timeStopping = TimeStopping(seconds = 30, verbose = true)
+        it.compile(
+            optimizer = Adam(clipGradient = ClipGradientByValue(0.1f)),
+            loss = Losses.SOFT_MAX_CROSS_ENTROPY_WITH_LOGITS,
+            metric = Metrics.ACCURACY,
+            callback = timeStopping,
```
zaleslaw, line 116: I have one idea here. Are you interested in the ticket about the support of multiple callbacks? At the current time, the API supports only one loss/metric/callback. From my point of view, multiple callbacks could make sense (stopping and logging to the CSV for example)

zaleslaw (approved):
I suppose it is ready to merge, but maybe you will be ready to refactor something related to the model usage.

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/callback/TimeStopping.kt
```diff
@@ -0,0 +1,47 @@
...
+ *
+ * This callback could be used to limit the training time of a model to a specified maximum duration.
+ * When the specified amount of time passes, the training of the model is stopped (if it has not already been
+ * finished or stopped) at the end of current epoch.
+ *
+ * @property [seconds] Maximum amount of time before stopping training.
+ * @property [verbose] Verbosity mode.
+ */
+public class TimeStopping(
+    private val seconds: Int = 86400,
```
zaleslaw, line 22: Add TODO here: upgrade to https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.time/-duration/ after upgrade to Kotlin 1.5 (1.4.30 at this moment)

examples/src/main/kotlin/examples/cnn/mnist/advanced/LeNetWithTimeStoppingCallback.kt
```diff
@@ -0,0 +1,129 @@
...
+private const val SEED = 12L
+private const val TEST_BATCH_SIZE = 1000
+
+/**
+ * This is an CNN based on an implementation of LeNet-5 from classic paper trained with TimeStopping callback.
+ *
+ * @see <a href="http://yann.lecun.com/exdb/publis/pdf/lecun-98.pdf">
+ *    Gradient-based learning applied to document recognition:[Yann LeCun, Léon Bottou, Yoshua Bengio, Patrick Haffner, 1998]</a>
+ */
+private val lenet5Classic = Sequential.of(
```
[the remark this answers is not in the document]
mkaze: Refactored all instances of LeNet5 using a builder function.

examples/src/main/kotlin/examples/cnn/mnist/advanced/LeNetWithTimeStoppingCallback.kt
mkaze: Yeah, sure that's a good idea. Actually, the Python API of Keras always expects a list of callbacks passed to the `fit` method. The same thing could be done here (of course, there is an irrelevant difference here: in KotlinDL, unlike Keras, the callbacks are passed to `compile` method instead of `fit`).

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/callback/TimeStopping.kt
mkaze: I added the TODO note; however, I am not sure if expecting and working with `Duration` has more benefit than simply using seconds value (from user perspective).

mkaze:
@zaleslaw I addressed your comments and made some changes. It's ready for another round of review.

examples/src/main/kotlin/examples/cnn/mnist/advanced/LeNetWithTimeStoppingCallback.kt
zaleslaw: Yeah, there is a difference in API due to internal implementation details.

PR #186, opened 2021-08-11
Fix typo in condition in random uniform initializer

juliabeliaeva:
"minVal" was checked the second time instead of "maxVal".


zaleslaw:
Could you please fix the typo in KDocs too, looks like minVal/maxVal should be exchanged

juliabeliaeva:
> Thanks for the fix! Could you please fix the typo in KDocs too, looks like minVal/maxVal should be exchanged @juliabeliaeva

Done: https://github.com/JetBrains/KotlinDL/pull/194

Issue #167, opened 2021-07-29
Add LiSHT activation function


zaleslaw:
The LiSHT (Non-Parametric Linearly Scaled Hyperbolic Tangent) activation function is missed in KotlinDL.

Read more about [LiSHT: Non-Parametric Linearly Scaled Hyperbolic Tangent Activation Function for Neural Networks](https://arxiv.org/abs/1901.05894)

The desired PR addressing this issue should include:

- [ ]  Implementation of activation class named as `LishtActivation` (you can take inspiration from the implementation of [`HardSigmoid`][1]  as reference) added to the [`Activations.kt`][2] file
- [ ]  Documentation of activation function
- [ ]  JUnit tests in `api ` module
- [ ]  Support for export of activation function to JSON (see [ModelSaver.kt][3])
- [ ]  Support for import of activation function from JSON (see [ModelLoader.kt][4])

[1]: https://github.com/JetBrains/KotlinDL/blob/master/api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/activation/Activations.kt#L344

[2]:
https://github.com/JetBrains/KotlinDL/blob/master/api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/activation/Activations.kt

[3]: https://github.com/JetBrains/KotlinDL/blob/master/api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/inference/keras/ModelSaver.kt

[4]: https://github.com/JetBrains/KotlinDL/blob/master/api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/inference/keras/ModelLoader.kt

P.S. If you want to take this ticket, please leave the comment below
P.P.S Read the [Contributing Guidelines](https://github.com/JetBrains/KotlinDL/blob/master/CONTRIBUTING.md).

The reference implementation could be taken from [tensorflow-addons](https://github.com/tensorflow/addons/blob/master/tensorflow_addons/activations/lisht.py)

PR #188, opened 2021-08-14
Added implementation of Non-Parametric Linearly Scaled Hyperbolic Tan…

knok16:
Added implementation of Non-Parametric Linearly Scaled Hyperbolic Tangent (LiSHT) Activation Function suggested in https://github.com/JetBrains/KotlinDL/issues/167

### DIFF
--- a/api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/inference/keras/KerasConstants.kt
+++ b/api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/inference/keras/KerasConstants.kt
@@ -86,7 +86,7 @@ internal const val INITIALIZER_TRUNCATED_NORMAL: String = "TruncatedNormal"
 internal const val INITIALIZER_CONSTANT: String = "Constant"
 internal const val INITIALIZER_VARIANCE_SCALING: String = "VarianceScaling"
 internal const val INITIALIZER_IDENTITY: String = "Identity"
-internal const val INITIALIZER_ORTHOGONAL:String = "Orthogonal"
+internal const val INITIALIZER_ORTHOGONAL: String = "Orthogonal"
 
 // Keras activations
 internal const val ACTIVATION_RELU: String = "relu"
@@ -105,6 +105,7 @@ internal const val ACTIVATION_LOG_SOFTMAX: String = "log_softmax"
 internal const val ACTIVATION_EXP: String = "exponential"
 internal const val ACTIVATION_MISH: String = "mish"
 internal const val ACTIVATION_HARDSHRINK: String = "hardshrink"
+internal const val ACTIVATION_LISHT: String = "lisht"
 
 // Layer settings
 internal const val CHANNELS_LAST: String = "channels_last"
--- a/api/src/test/kotlin/org/jetbrains/kotlinx/dl/api/core/activation/ActivationTest.kt
+++ b/api/src/test/kotlin/org/jetbrains/kotlinx/dl/api/core/activation/ActivationTest.kt
@@ -31,4 +31,16 @@ open class ActivationTest {
             )
         }
     }
+
+    protected fun Activation.apply(input: FloatArray): FloatArray = EagerSession.create().use { session ->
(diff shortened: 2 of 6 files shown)


api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/inference/keras/KerasConstants.kt
```diff
@@ -105,6 +105,7 @@ internal const val ACTIVATION_LOG_SOFTMAX: String = "log_softmax"
 internal const val ACTIVATION_EXP: String = "exponential"
 internal const val ACTIVATION_MISH: String = "mish"
 internal const val ACTIVATION_HARDSHRINK: String = "hardshrink"
+internal const val ACTIVATION_LISHT: String = "lisht"
```
knok16, line 108: As I understand it should be equals to function name here: https://github.com/tensorflow/addons/blob/master/tensorflow_addons/activations/lisht.py#L22

api/src/test/kotlin/org/jetbrains/kotlinx/dl/api/core/activation/ActivationTest.kt
```diff
@@ -31,4 +31,16 @@ open class ActivationTest {
...
+
+    protected fun Activation.apply(input: FloatArray): FloatArray = EagerSession.create().use { session ->
+        val tf = Ops.create(session)
+
+        val operand = this.apply(tf, tf.constant(input))
+
+        operand.asOutput().tensor().copyTo(FloatArray(input.size))
+    }
+
+    protected fun assertActivationFunction(activation: Activation, input: FloatArray, expected: FloatArray) {
```
knok16, line 43: What do you think about `assertActivationFunction` without `actual` parameter - as I see in most (all?) cases the caller just pass an array of zeroes, which can be prepared inside `assertActivationFunction`?

api/src/test/kotlin/org/jetbrains/kotlinx/dl/api/core/activation/ActivationTest.kt
```diff
@@ -31,4 +31,16 @@ open class ActivationTest {
             )
         }
     }
+
+    protected fun Activation.apply(input: FloatArray): FloatArray = EagerSession.create().use { session ->
```
knok16, line 35: Split `assertActivationFunction` into 2 blocks for easier composition (in case someone wants to assert results differently):
1. apply activation to input and produce output
2. actually assert result

WDYT?

knok16:
Run tests locally: 213 of 213 passed

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/inference/keras/KerasConstants.kt
zaleslaw: I suppose yes, it should work for serialization/deserialization needs

api/src/test/kotlin/org/jetbrains/kotlinx/dl/api/core/activation/ActivationTest.kt
zaleslaw: Yeah, you are absolutely right about the `actual` parameter, it's just an empty array for copying needs

zaleslaw:
So, I'd like this PR and your ideas about minor refactoring in tests!

api/src/test/kotlin/org/jetbrains/kotlinx/dl/api/core/activation/ActivationTest.kt
zaleslaw: it's an interesting and useful idea (with emulation of apply function on EagerSession). Hope it will be widely used in tests for primitives: initializers, layers, and so on.

zaleslaw:
Hey @knok16 your PR is merged, thanks for the contribution, I'm going to create a few new ticket at the next week, but feel free to propose something from your side via issue or start a discussion

PR #209, opened 2021-09-01
Refactor image preprocessing operations to allow new operations addition

### LINKED ISSUE #208: Image preprocessing issues
1. Adding new image preprocessing operations is difficult as it requires adding new fields to `ImagePreprocessing` class and adjusting `Preprocessing#finalShape` and `Preprocessing#handleFile`.
2. Order in which operations are defined is not respected.

juliabeliaeva:
Fixes #208.

### DIFF
--- a/dataset/src/main/kotlin/org/jetbrains/kotlinx/dl/dataset/preprocessor/Preprocessing.kt
+++ b/dataset/src/main/kotlin/org/jetbrains/kotlinx/dl/dataset/preprocessor/Preprocessing.kt
@@ -32,8 +32,10 @@ public class Preprocessing {
     public val finalShape: ImageShape
         get() {
             var imageShape = if (::load.isInitialized) load.imageShape else null
-            for (operation in imagePreprocessingStage.operations) {
-                imageShape = operation.getOutputShape(imageShape)
+            if (::imagePreprocessingStage.isInitialized) {
+                for (operation in imagePreprocessingStage.operations) {
+                    imageShape = operation.getOutputShape(imageShape)
+                }
             }
             if (imageShape == null) {
                 throw IllegalStateException(
@@ -55,12 +57,14 @@ public class Preprocessing {
     internal fun handleFile(file: File): Pair<FloatArray, ImageShape> {
         var image = load.fileToImage(file)
 
-        for (operation in imagePreprocessingStage.operations) {
-            if (operation is Save) {
-                operation.imageToFile(file.name, image)
-                continue
+        if (::imagePreprocessingStage.isInitialized) {
+            for (operation in imagePreprocessingStage.operations) {
+                if (operation is Save) {
+                    operation.imageToFile(file.name, image)
+                    continue
+                }
+                image = operation.apply(image)
             }
-            image = operation.apply(image)
         }
 
         var tensor = OnHeapDataset.toRawVector(imageToByteArray(image, load.colorMode))
--- a/dataset/src/main/kotlin/org/jetbrains/kotlinx/dl/dataset/preprocessor/Preprocessing.kt
+++ b/dataset/src/main/kotlin/org/jetbrains/kotlinx/dl/dataset/preprocessor/Preprocessing.kt
@@ -8,8 +8,8 @@ package org.jetbrains.kotlinx.dl.dataset.preprocessor
 import org.jetbrains.kotlinx.dl.dataset.OnHeapDataset
 import org.jetbrains.kotlinx.dl.dataset.image.ImageConverter.Companion.imageToByteArray
 import org.jetbrains.kotlinx.dl.dataset.preprocessor.image.ImagePreprocessing
+import org.jetbrains.kotlinx.dl.dataset.preprocessor.image.ImagePreprocessorBase
 import org.jetbrains.kotlinx.dl.dataset.preprocessor.image.Loading
-import org.jetbrains.kotlinx.dl.dataset.preprocessor.image.Save
 import java.awt.image.BufferedImage
 import java.io.File
 
@@ -59,11 +59,8 @@ public class Preprocessing {
 
         if (::imagePreprocessingStage.isInitialized) {
             for (operation in imagePreprocessingStage.operations) {
-                if (operation is Save) {
-                    operation.imageToFile(file.name, image)
-                    continue
-                }
                 image = operation.apply(image)
+                (operation as? ImagePreprocessorBase)?.save?.save(file.name, image)
             }
         }
 
--- a/dataset/src/main/kotlin/org/jetbrains/kotlinx/dl/dataset/preprocessor/Preprocessing.kt
+++ b/dataset/src/main/kotlin/org/jetbrains/kotlinx/dl/dataset/preprocessor/Preprocessing.kt
@@ -8,6 +8,7 @@ package org.jetbrains.kotlinx.dl.dataset.preprocessor
 import org.jetbrains.kotlinx.dl.dataset.OnHeapDataset
 import org.jetbrains.kotlinx.dl.dataset.image.ImageConverter.Companion.imageToByteArray
 import org.jetbrains.kotlinx.dl.dataset.preprocessor.image.ImagePreprocessing
+import org.jetbrains.kotlinx.dl.dataset.preprocessor.image.Loading
 import org.jetbrains.kotlinx.dl.dataset.preprocessor.image.Save
 import java.awt.image.BufferedImage
 import java.io.File
@@ -18,6 +19,9 @@ import java.io.File
  * Could be used to handle directory of images or one image file.
  */
 public class Preprocessing {
+    /** */
+    public lateinit var load: Loading
+
     /** This stage describes the process of image loading and transformation before converting to tensor. */
     public lateinit var imagePreprocessingStage: ImagePreprocessing
 
@@ -27,7 +31,7 @@ public class Preprocessing {
     /** Returns the final shape of data when image preprocessing is applied to the image. */
     public val finalShape: ImageShape
         get() {
-            var imageShape = if (imagePreprocessingStage.isLoadInitialized) imagePreprocessingStage.load.imageShape else null
+            var imageShape = if (::load.isInitialized) load.imageShape else null
             for (operation in imagePreprocessingStage.operations) {
                 imageShape = operation.getOutputShape(imageShape)
             }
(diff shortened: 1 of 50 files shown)


dataset/src/test/kotlin/org/jetbrains/kotlinx/dl/dataset/preprocessor/PreprocessingFinalShapeTest.kt
```diff
@@ -0,0 +1,134 @@
+package org.jetbrains.kotlinx.dl.dataset.preprocessor
+
+import org.jetbrains.kotlinx.dl.dataset.image.ColorOrder
+import org.jetbrains.kotlinx.dl.dataset.preprocessor.image.*
+import org.junit.jupiter.api.Assertions.assertEquals
+import org.junit.jupiter.api.Test
+import java.io.File
+
+class PreprocessingFinalShapeTest {
```
zaleslaw, line 9: What do you think about adding tests for comparing preprocessed images with earlier preprocessed and saved to the resources? Is it too difficult to add?

dataset/src/main/kotlin/org/jetbrains/kotlinx/dl/dataset/preprocessor/image/Resize.kt
```diff
@@ -33,7 +33,12 @@ public class Resize(
     public var renderingSpeed: RenderingSpeed = RenderingSpeed.MEDIUM,
     public var enableAntialiasing: Boolean = true
 ) : ImagePreprocessor {
-    override fun apply(image: BufferedImage, inputShape: ImageShape): Pair<BufferedImage, ImageShape> {
+
+    override fun getOutputShape(inputShape: ImageShape?): ImageShape {
+        return ImageShape(outputWidth.toLong(), outputHeight.toLong(), 3)
```
zaleslaw, line 38: Could it be applied to the 1-channel images?

dataset/src/main/kotlin/org/jetbrains/kotlinx/dl/dataset/preprocessor/image/Loading.kt
```diff
@@ -16,7 +16,7 @@ import java.nio.file.Path
 import kotlin.streams.toList
 
 /**
- * This image preprocessor defines the Loading operation.
+ * This class defines the Loading operation.
```
zaleslaw, line 19: Highly likely it should be moved to the first-level in the DSL like

   loadFile {
                    pathToData = File("test.jpg")
                    imageShape = ImageShape(20, 20, 3)
                    colorMode = ColorOrder.BGR
                }
transformImage {
                resize {
                    outputWidth = 100
                    outputHeight = 100
                    interpolation = InterpolationType.NEAREST
                }
            }

And also we could add loadFile/loadDirectories ops or 
load {
      file {
      }
}

load {
      directory {
      }
}

load {
      database {
      }
}

zaleslaw: But it could be a following ticket of course, it's not a part of this refactoring

dataset/src/main/kotlin/org/jetbrains/kotlinx/dl/dataset/preprocessor/image/ImagePreprocessor.kt
```diff
@@ -11,13 +11,19 @@ import java.awt.image.BufferedImage
...
-     * Transforms [image] with [inputShape] to the new image with the new shape.
+     * Computes output image shape for the provided [inputShape].
+     * @param inputShape image input shape. Null value means that input image size is not known.
+     *                   This is useful for operations with a fixed output size,
+     *                   which should return a non-null value in this case.
      *
-     * @return Pair <new image; new shape>.
+     * @return output image shape
      */
-    public fun apply(image: BufferedImage, inputShape: ImageShape): Pair<BufferedImage, ImageShape>
```
zaleslaw, line 18: So, I had yet two ideas, but didn't implement yet
1) return pair, but calculate outputShape by inputShape for each Preprocessor
2) return new object of class OurGreatImage (with fields BufferedImage and ImageShape), later it's easy to add these prepocessors like extension functions to the OurGreatImage for example

Let's discuss both approaches?

dataset/src/main/kotlin/org/jetbrains/kotlinx/dl/dataset/preprocessor/image/ImagePreprocessing.kt
```diff
@@ -21,37 +20,16 @@ public class ImagePreprocessing {
...
         get() = ::load.isInitialized
 
-    /** True, if [save] is initialized. */
-    public val isSaveInitialized: Boolean
-        get() = ::save.isInitialized
+    /** */
+    internal val operations = mutableListOf<ImagePreprocessor>()
+
+    public fun addOperation(operation: ImagePreprocessor) {
+        operations.add(operation)
```
zaleslaw, line 31: Yeah, thanks, I'd really like to see the removal of these functions, no needs now

dataset/src/main/kotlin/org/jetbrains/kotlinx/dl/dataset/preprocessor/image/Cropping.kt
```diff
@@ -29,22 +30,27 @@ public class Cropping(
...
 ) : ImagePreprocessor {
-    override fun apply(image: BufferedImage, inputShape: ImageShape): Pair<BufferedImage, ImageShape> {
-        val croppedImageShape =
-            ImageShape(
-                width = inputShape.width!! - left - right,
-                height = inputShape.height!! - top - bottom,
-                channels = inputShape.channels
-            )
+
+    override fun getOutputShape(inputShape: ImageShape?): ImageShape? {
```
zaleslaw, line 34: Looks like all transformation should have implementation and ability to return/control shape.

dataset/src/main/kotlin/org/jetbrains/kotlinx/dl/dataset/preprocessor/Preprocessing.kt
```diff
@@ -51,45 +49,25 @@ public class Preprocessing {
...
-        if (imagePreprocessingStage.isResizeInitialized) {
-            val (resizedImage, resizedShape) = imagePreprocessingStage.resize.apply(image, shape)
-            image = resizedImage
-            shape = resizedShape
-        }
-
-        if (imagePreprocessingStage.isSaveInitialized) {
-            imagePreprocessingStage.save.imageToFile(filename, image, shape)
+        for (operation in imagePreprocessingStage.operations) {
+            if (operation is Save) {
```
zaleslaw, line 55: Looks like we have two sort of transformation: intermediate and start/final.
What do you think about Save op? Is it an imageTransformation or we should have a separate block for it at the top level?
load {
}
transformImage {
}
save {
}

zaleslaw: But on other side, what if we want to save checkpoints after each imageTransformation (or display it)

zaleslaw (changes requested):
So, I'd really like your job in this PR; I wrote a few notes for you and suggested discussing it.

dataset/src/main/kotlin/org/jetbrains/kotlinx/dl/dataset/preprocessor/image/Resize.kt
juliabeliaeva: I think so, however the output will have 3 channels anyway. Ideally we'd want to keep the number of channels in the image. So `Resize#getOutputShape` should return `ImageShape(outputWidth.toLong(), outputHeight.toLong(), null)` (we need to make `ImageShape#channels` nullable for that).

dataset/src/main/kotlin/org/jetbrains/kotlinx/dl/dataset/preprocessor/Preprocessing.kt
juliabeliaeva: > But on other side, what if we want to save checkpoints after each imageTransformation (or display it)

Yes, that what I was thinking.

One of the possible ways it could look like:

```
resize {
    outputWidth = 400
    outputHeight = 400
    interpolation = InterpolationType.NEAREST
    save {
        dirLocation = preprocessedImagesDirectory
    }
}
```

dataset/src/main/kotlin/org/jetbrains/kotlinx/dl/dataset/preprocessor/image/ImagePreprocessor.kt
juliabeliaeva: Could you explain why a separate shape is needed? `BufferedImage` has this information already, so it seems a bit redundant to me. Not to mention that because of the nullable fields using `ImageShape` might be awkward in this case.

dataset/src/main/kotlin/org/jetbrains/kotlinx/dl/dataset/preprocessor/image/Save.kt
```diff
@@ -24,7 +24,7 @@ public class Save(
     public var dirLocation: File? = null
 ) : ImagePreprocessor {
     @Throws(IOException::class)
-    internal fun imageToFile(filename: String, image: BufferedImage, shape: ImageShape): File {
+    internal fun imageToFile(filename: String, image: BufferedImage): File {
         val outputFile: File = if (dirLocation!!.isDirectory) {
             File("${dirLocation}\\${filename}")
```
YokiToki, line 29: How about fix this line `File.separator`? This is not correct for unix systems now.

zaleslaw: Many thanks for the comment, do you have any suggestions for quick fix?

YokiToki: I think there is a problem with declaring dirLocation as nullable. Passing null will just throw a NullPointerException. And I want to note that if the value of `dirLocation` is not a directory, then each processing result will overwrite the file.

juliabeliaeva: > How about fix this line with File.separator? This is not correct for unix systems now.

I think this fix should be done as a separate pull request.

juliabeliaeva:
I pushed some fixes for this PR:
1. Defined output shapes for all operations.
2. Moved `load` out of `transformImage`. I did not split it to `loadFile/loadDirectory` as it turned into another major refactoring and I think it should be done as a separate PR.
3. Moved `save` to individual operations.
4. Added a couple of tests. I thought that perhaps adding more images to the repository may not be optimal, so I've written some simple in-memory tests.

juliabeliaeva:
Rebased on latest master to resolved conflicts.

dataset/src/main/kotlin/org/jetbrains/kotlinx/dl/dataset/preprocessor/image/Save.kt
zaleslaw: @YokiToki we are going to split into two parts the logic of loading the whole directory and one file, and thanks for the note about the Unix system and possible overwrinting

PR #210, opened 2021-09-01
Ignore "build" directories

juliabeliaeva:


zaleslaw:
Cool refactoring, don't think this way

PR #214, opened 2021-09-06
replaced custom stack implementation by the one from standart library

knok16:


zaleslaw:
I didn't find Stack implementations in Kotlin Collections but found the same notes that it's legal to implement it via mutable list.
I'm trying to avoid Java Stack for many years due to legacy history in the codebase. This is a reason why the current implementations exist. It's not a re-inventing the wheel, it's just a simple helper for one usage

knok16:
got it, thanks for the explanation. let's then reject current PR.

PR #218, opened 2021-09-09
reduced time complexity of FloatArray::argmax to linear

knok16:
Reduced time complexity of argmax method from O(n^2) to linear, also, since it is now a generic function, make it work with empty arrays


zaleslaw:
Thanks, @knok16. It's a really awesome fix; correct me if I'm wrong. It reduced from N^2 to 2N (N for maxOrNull and N in the bad case for the indexOfFirst)

knok16:
Yes, it is O(n) now, with two passes through the array. It also can be rewritten to something like `this.asSequence().withIndex().maxByOrNull { it.value }?.index ?: -1` - to have a single pass which can be beneficial in some cases (I believe for cases if you have a very long array), but I think constant for such approach would be comparable to the provided solution.

PR #222, opened 2021-09-15
[WIP] Added ParametrizedLayer and TrainableLayer interfaces

knok16:
Added `ParametrizedLayer` and `TrainableLayer` interfaces, moved `paramCount` and `isTrainable` fields there from `Layer,` also added `VariableDto` data class to encapsulate all necessary info about TensorFlow `Variable` (like name, shape, initializer, and regularizer), exposed it in `ParametrizedLayer`

### DIFF
--- a/api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/GraphTrainableModel.kt
+++ b/api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/GraphTrainableModel.kt
@@ -10,8 +10,7 @@ import mu.KotlinLogging
 import org.jetbrains.kotlinx.dl.api.core.callback.Callback
 import org.jetbrains.kotlinx.dl.api.core.exception.RepeatableLayerNameException
 import org.jetbrains.kotlinx.dl.api.core.history.*
-import org.jetbrains.kotlinx.dl.api.core.layer.Layer
-import org.jetbrains.kotlinx.dl.api.core.layer.NoGradients
+import org.jetbrains.kotlinx.dl.api.core.layer.*
 import org.jetbrains.kotlinx.dl.api.core.layer.core.ActivationLayer
 import org.jetbrains.kotlinx.dl.api.core.layer.core.Dense
 import org.jetbrains.kotlinx.dl.api.core.layer.core.Input
@@ -54,7 +53,9 @@ public abstract class GraphTrainableModel(vararg layers: Layer) : TrainableModel
     public val logger: KLogger = KotlinLogging.logger {}
 
     /** The layers to describe the model design. Main part of the internal state of the model. */
-    public var layers: List<Layer> = listOf(*layers)
+    init {
+        this.layers = listOf(*layers)
+    }
 
     /** First layer that is responsible for the input shape of the Neural Network. */
     public val inputLayer: Input
@@ -178,7 +179,7 @@ public abstract class GraphTrainableModel(vararg layers: Layer) : TrainableModel
 
         yPredOp = forward(xOp, inputLayer)
         lossOp = buildLossFunction(loss)
-        targets = optimizer.prepareTargets(kGraph, tf, lossOp)
+        targets = optimizer.prepareTargets(kGraph, trainableLayerVariables().map { it.variable }, tf, lossOp)
 
         predictionOp = when (loss) {
             is SoftmaxCrossEntropyWithLogits -> tf.withName(OUTPUT_NAME).nn.softmax(yPredOp)
@@ -191,14 +192,13 @@ public abstract class GraphTrainableModel(vararg layers: Layer) : TrainableModel
 
     private fun buildLossFunction(loss: LossFunction): Operand<Float> {
         val basicLoss = loss.apply(tf, yPredOp, yTrueOp, numberOfLossesOp)
-        var totalLoss = basicLoss
         // TODO: probably regularization output should be divided on numberOfLossesOp and changed together with loss before averaging
-        kGraph.variableRegularizers.forEach { (variable, regularizer) ->
-            run {
-                totalLoss = tf.math.add(totalLoss, regularizer.apply(tf, variable))
+        val result = trainableLayerVariables()
+            .mapNotNull { it.regularizer?.let { r -> it.variable to r } }
+            .fold(basicLoss) { acc, (variable, regularizer) ->
+                tf.math.add(acc, regularizer.apply(tf, variable))
             }
-        }
-        return tf.withName(TRAINING_LOSS).identity(totalLoss)
+        return tf.withName(TRAINING_LOSS).identity(result)
     }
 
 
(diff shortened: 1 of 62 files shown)


api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/TrainableLayer.kt
```diff
@@ -0,0 +1,32 @@
+package org.jetbrains.kotlinx.dl.api.core.layer
+
+import org.tensorflow.op.core.Variable
+
+public interface TrainableLayer : ParametrizedLayer {
```
knok16, line 5: added ParametrizedLayer/TrainableLayer interfaces, so only the layers that actually have parameters and/or can be trained will extend it

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/VariableDto.kt
```diff
@@ -0,0 +1,38 @@
...
+
+import org.jetbrains.kotlinx.dl.api.core.initializer.Initializer
+import org.jetbrains.kotlinx.dl.api.core.regularizer.Regularizer
+import org.jetbrains.kotlinx.dl.api.core.util.getDType
+import org.tensorflow.Operand
+import org.tensorflow.Shape
+import org.tensorflow.op.Ops
+import org.tensorflow.op.core.Variable
+
+public data class VariableDto(
```
knok16, line 11: Added class to encapsulate all related to TensorFlow's Variable<Float> info, which is exposed by `ParametrizedLayer::variables` and no need to pass variable initializers and regularizers through `KGraph`. The main idea here is that layers already represent computation graph and all related info can be stored here

knok16: Because of it, it is possible to remove `KGraph` argument from `Layer::build` method

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/inference/TensorFlowInferenceModel.kt
```diff
@@ -28,6 +31,9 @@ import java.util.*
  * Provides functionality to make predictions and model loading.
  */
 public open class TensorFlowInferenceModel : InferenceModel() {
+    /** The layers to describe the model design. Main part of the internal state of the model. */
+    public lateinit var layers: List<Layer>
```
knok16, line 35: don't like this change, to get access to refactored data in `isOptimizerNameAndRelatedToFrozenLayer` and `getVariablesAndTensors` layers required, will think about it

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/inference/keras/WeightLoader.kt
```diff
@@ -197,7 +199,7 @@ private fun fillConv2DVariablesFromKeras(
             }
             "bias:0" -> {
                 val biasVariableName = convBiasVarName(layerName, dim = 2)
-                val biasShape = (model.getLayer(layerName) as Conv2D).biasShapeArray
+                val biasShape = (model.getLayer(layerName) as Conv2D).biasShapeArray!!
```
knok16, line 202: added workaround like it is `fillBatchNormVariablesFromKeras`, probably better to remove these `!!` and just don't fill layers if they don't have bias variable.

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/GraphTrainableModel.kt
```diff
@@ -176,7 +177,7 @@ public abstract class GraphTrainableModel(vararg layers: Layer) : TrainableModel
 
         yPredOp = forward(xOp, inputLayer)
         lossOp = buildLossFunction(loss)
-        targets = optimizer.prepareTargets(kGraph, tf, lossOp)
+        targets = optimizer.prepareTargets(kGraph, layers.trainableLayerVariables().map { it.variable }, tf, lossOp)
```
knok16, line 180: pass TensorFlow variables directly to optimizer, instead of implicitly through `KGraph`

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/convolutional/AbstractConv.kt
```diff
@@ -62,40 +59,53 @@ public abstract class AbstractConv(
...
-        kernelShape = computeKernelShape(numberOfChannels)
-        biasShape = computeBiasShape(numberOfChannels)
-
         // should be calculated before addWeight because it's used in calculation
         val inputDepth = numberOfChannels // number of input channels
         val outputDepth = getOutputDepth(numberOfChannels) // number of output channels
-        fanIn = (inputDepth * multiply(*kernelSizeInternal)).toInt()
-        fanOut = ((outputDepth * multiply(*kernelSizeInternal)).toDouble() /
-                multiply(*stridesInternal).toDouble()).roundToInt()
+        fanIn = (inputDepth * multiply(kernelSizeInternal)).toInt()
```
knok16, line 82: `fanIn`/`fanOut` used only for preparing initializers, and filled only for some of the layers, maybe make sense to delete it from `Layer`

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/ParametrizedLayer.kt
```diff
@@ -0,0 +1,39 @@
...
+
+import org.jetbrains.kotlinx.dl.api.core.shape.numElements
+import org.tensorflow.Session
+import org.tensorflow.op.core.Variable
+
+public interface ParametrizedLayer {
+    /**
+     * Layer's variables
+     */
+    public val variables: List<VariableDto>
```
knok16, line 11: By having it on layer explicitly, `weight` setter/getter can be implemented as an extension function outside of the `Layer` to not clutter its code:
```
/**
 *  Layer's weights
 *  Require parent model to be set on layer
 */
public var Layer.weights: Map<String, Array<*>>
    get() {
        if (this !is ParametrizedLayer) return emptyMap()

        val model = parentModel
        requireNotNull(model) { "Layer '$name' is not related to any model" }

        val variablesOrder = variables
        val runner = model.session.runner()
        variablesOrder.map { it.variable }.forEach(runner::fetch)
        val weights = runner.run().map { it.convertTensorToMultiDimArray() }

        return variablesOrder.map { it.name }.zip(weights).toMap()
    }
    set(weights) {
        if (this !is ParametrizedLayer) return

        val model = parentModel
        requireNotNull(model) { "Layer '$name' is not related to any model" }

        for ((name, value) in weights) {
            model.assignVariable(name, value)
        }
    }
```

and removed from all of the layers (in most of the cases it is simple 
``` 
override var weights: Map<String, Array<*>>
        get() = emptyMap()
        set(value) = assignWeights(value)
```
), along with `Layer::extractWeights`/`Layer::assignWeights`.

PR #225, opened 2021-09-19
Remove redundant array initializers

juliabeliaeva:
There is a number of places in the code where the following pattern is used:
```
val array = Array(length) { /*initializer*/ }
for ( i in 0 until length) {
  array[i] = /*other initializer*/
}
```
Sometimes the first initializer creates another array, which is then thrown out in the `for`-cycle. 
This commit fixes such places in the code.


zaleslaw:
Yeah, there are a few places in the ancient code. When I made the first steps with the Kotlin Arrays and code practices, thanks for fixing it, I just blindly copy-pasted the old code into the new one.

PR #229, opened 2021-09-20
Close input stream after reading

### LINKED ISSUE #228: `ImageConverter` and `Loading` do not close opened streams
See:
https://github.com/JetBrains/KotlinDL/blob/540bf0316d38d6216ef99b8afd8403cdd72e8889/dataset/src/main/kotlin/org/jetbrains/kotlinx/dl/dataset/preprocessor/image/Loading.kt#L40
https://github.com/JetBrains/KotlinDL/blob/540bf0316d38d6216ef99b8afd8403cdd72e8889/dataset/src/main/kotlin/org/jetbrains/kotlinx/dl/dataset/image/ImageConverter.kt#L31
https://github.com/JetBrains/KotlinDL/blob/540bf0316d38d6216ef99b8afd8403cdd72e8889/dataset/src/main/kotlin/org/jetbrains/kotlinx/dl/dataset/image/ImageConverter.kt#L48

juliabeliaeva:
Fixes #228.


zaleslaw:
Great, thanks for finding a bug!

PR #232, opened 2021-09-20
Fix property has been uninitialized issue in dense layer without bias

knok16:


zaleslaw:
FunctionalModelTest.summary due to changes in the static graph, I suppose due to this bug fix

 at org.junit.jupiter.api.AssertionUtils.fail(AssertionUtils.java:55)
  at org.junit.jupiter.api.AssertionUtils.failNotEqual(AssertionUtils.java:62)
  at org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:182)
  at org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:177)
  at org.junit.jupiter.api.Assertions.assertEquals(Assertions.java:1124)
  at org.jetbrains.kotlinx.dl.api.core.models.FunctionalModelTest.summary(FunctionalModelTest.kt:97)
  at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
  at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
  at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
  at java.lang.reflect.Method.invoke(Method.java:498)

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/core/Dense.kt
```diff
@@ -117,19 +119,22 @@ public class Dense(
         set(value) = assignWeights(value)
 
     private fun extractDenseWeights(): Map<String, Array<*>> {
```
knok16, line 121: I checked other layers with bias parameter (convolution ones), and looks like they have similar problems:
- `weights` getter will try to extract bias parameters even though it is not enabled for the layer
- `paramCount` includes `bias` even if `useBias=false` has been used

api/src/test/kotlin/org/jetbrains/kotlinx/dl/api/core/models/FunctionalModelTest.kt
```diff
@@ -93,7 +93,7 @@ internal class FunctionalModelTest {
             correctTestModel.summary()
         )
 
-        val graphLines = correctTestModel.kGraph().toString().split('\n').toList()
+        val graphLines = correctTestModel.kGraph().toString().split('\n').toSet()
```
zaleslaw, line 96: Does this change fix the test? I thought the problem is in the graph operators

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/core/Dense.kt
zaleslaw: Yes, this is a typical pattern for all layers with the useBias flag. It looks like it requires common refactoring.
Could you please do it in this PR? It will be very helpful if you do it today or tomorrow. Do you have any time for that?

api/src/test/kotlin/org/jetbrains/kotlinx/dl/api/core/models/FunctionalModelTest.kt
knok16: Yes, the test is green. I think we should not rely on ordering here, since internally `toString()` uses `org.tensorflow.Graph::operations` which states:
> The order of iteration is unspecified

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/core/Dense.kt
zaleslaw: Let's keep this idea for the future, it involves a lot of refactoring in the master branch which should become a release branch in two days. I suggest to finish this ticket without proposed DTO object

api/src/test/kotlin/org/jetbrains/kotlinx/dl/api/core/models/FunctionalModelTest.kt
zaleslaw: Hmm, interesting that it was broken on this test, which could influence on a graph

zaleslaw:
Hi, sorry for the question, do you have any time to finish it today or tomorrow? @knok16 It will be really helpful due to 0.3 release

knok16:
Done, please, review

zaleslaw:
Great, thanks for the fast fix, now it merged

PR #239, opened 2021-09-24
[WIP] Add basic tests for ONNX module

femialaka:
This is a draft PR for the Tests.

Some ONNX Models were not complete e.g Lenet, so I used ResNet instead.

The Transpose class was also incomplete so I could not Test it.

I would like to know how do you generate input data for the Transpose class.

I will appreciate your feedback on the tests and anything that I may add, thanks


zaleslaw:
I'll do the review this week I comment the proposed draft

zaleslaw:
It's outdated and probably we don't need these tests for now

PR #249, opened 2021-09-27
Generalize reshape layer to higher dimensions

knok16:


api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/reshaping/Reshape.kt
```diff
@@ -36,47 +36,18 @@ public class Reshape(
...
-                )
-            )
-            else -> throw UnsupportedOperationException("Reshaping for ${targetShape.size} is not supported yet!")
-        }
+
+        units = tf.constant(IntArray(targetShape.size + 1) {
+            if (it == 0) -1 else targetShape[it - 1]
+        })
 
         fanIn = tensorShape.numElements().toInt()
```
knok16, line 44: Should it even be populated?

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/reshaping/Reshape.kt
```diff
@@ -36,47 +36,18 @@ public class Reshape(
     override fun build(tf: Ops, kGraph: KGraph, inputShape: Shape) {
         val tensorShape = TensorShape(inputShape)
         val amountOfNeuronsInFlattenLayer = (tensorShape.numElements() / abs(tensorShape.size(0))).toInt()
```
knok16, line 38: `amountOfNeuronsInFlattenLayer` looks suspicious

zaleslaw: self-documented variable :)

PR #253, opened 2021-10-02
Onnx Yolo v4 example

kokorins:
I have some doubts in the approach I'm using.  Though classes seems to be ok, but I cant interpret the bounding box coordinates.


zaleslaw:
Hi @kokorins happy to see you again

First of all, I need to share with you the original [model](https://github.com/onnx/models/tree/master/vision/object_detection_segmentation/yolov4) and pre and post-processing for it.

Looks like you use the simpler version of postprocessing without any maths related to the intersection of boxes and so on (probably it could be missed).

Also, need to note that different YOLO versions have different output tensors and structures (v2 and v4 for example).

But what is the current problem with your numbers?

PR #257, opened 2021-10-10
Remove CustomPreprocessor

juliabeliaeva:
Since the refactoring in 540bf0316d38d6216ef99b8afd8403cdd72e8889 any number of preprocessors of any kind could be used, so there is no need to have a special "custom" one.


zaleslaw:
Nice point!

PR #278, opened 2021-11-15
Simplify layer variable names definition

juliabeliaeva:


api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/convolutional/SeparableConv2D.kt
```diff
@@ -122,17 +116,8 @@ public class SeparableConv2D(
...
-
-    private fun definePointwiseKernelVariableName(): String =
-        if (name.isNotEmpty()) separableConv2dPointwiseKernelVarName(name) else POINTWISE_KERNEL_VARIABLE_NAME
-
-    private fun defineBiasVariableName(): String =
-        if (name.isNotEmpty()) separableConv2dBiasVarName(name) else BIAS_VARIABLE_NAME
-
     private fun createSeparableConv2DVariables(tf: Ops, kGraph: KGraph) {
-        val depthwiseKernelVariableName = defineDepthwiseKernelVariableName()
+        val depthwiseKernelVariableName = separableConv2dDepthwiseKernelVarName(name)
```
zaleslaw, line 120: why does depthwise was changed on separable?

zaleslaw:
I'm ok with that change

Issue #170, opened 2021-07-29
Add Soft shrink activation function


zaleslaw:
The Soft shrink activation function is missed in KotlinDL.

The desired PR addressing this issue should include:

- [ ]  Implementation of activation class named as `SoftShrinkActivation` (you can take inspiration from the implementation of [`HardSigmoid`][1]  as reference) added to the [`Activations.kt`][2] file
- [ ]  Documentation of activation function
- [ ]  JUnit tests in `api ` module
- [ ]  Support for export of activation function to JSON (see [ModelSaver.kt][3])
- [ ]  Support for import of activation function from JSON (see [ModelLoader.kt][4])

[1]: https://github.com/JetBrains/KotlinDL/blob/master/api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/activation/Activations.kt#L344

[2]:
https://github.com/JetBrains/KotlinDL/blob/master/api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/activation/Activations.kt

[3]: https://github.com/JetBrains/KotlinDL/blob/master/api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/inference/keras/ModelSaver.kt

[4]: https://github.com/JetBrains/KotlinDL/blob/master/api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/inference/keras/ModelLoader.kt

P.S. If you want to take this ticket, please leave the comment below
P.P.S Read the [Contributing Guidelines](https://github.com/JetBrains/KotlinDL/blob/master/CONTRIBUTING.md).

The reference implementation could be taken from [tensorflow-addons](https://github.com/tensorflow/addons/blob/master/tensorflow_addons/activations/softshrink.py)

michalharakal:
Hello @zaleslaw, I would like to implement this one.

zaleslaw:
Great @michalharakal will wait a PR!

zaleslaw:
Hi, @michalharakal Do you have any updates on this issue? Do you need some help from my side?

michalharakal:
@zaleslaw, sorry for delay, I was just busy with other stuff, by accident I could gave on a Kotlin related conference as backup. I plan to have a look on issue this weekend. 
In case, this issue is time critical, feel free to reassigne it, I will pick another one later, if I have more time.

zaleslaw:
[...]

PR #286, opened 2021-11-30
 Add Soft shrink activation function #170 

michalharakal:


zaleslaw:
Thanks for the PR @michalharakal 
Now it's a part of KotlinDL!

PR #315, opened 2021-12-20
Add transpose convolution layers

### LINKED ISSUE #124: Add Conv1DTranspose, Conv2DTranspose, Conv3DTranspose layers
Currently, the support for `Conv2DTranspose` and `Conv3DTranspose`, `Conv1DTranspose`  are missed, and it would be great to add support for these layers. The desired PR addressing this issue should include:

- [ ]  Implementation of layer class named as `Conv2DTranspose` (you can take inspiration from the implementation of [`Conv2D`][1] as reference)
- [ ]  Implementation of layer class named as `Conv3DTranspose ` (you can take inspiration from the implementation of [`Conv3D`][4] as reference)
- [ ]  Implementation of layer class named as `Conv1DTranspose ` 
- [ ] Common hierarchy of all ConvTranspose layers with abstract class with the common functionality
- [ ]  Documentation of layer and all non-private methods
- [ ]   JUnit tests in `api ` module
- [ ]   Support for export of layer to JSON (see [ModelSaver.kt][2])
- [ ]   Support for import of layer from JSON (see [ModelLoader.kt][3])

This operation is sometimes called "deconvolution" after [Deconvolutional Networks](http://www.matthewzeiler.com/pubs/cvpr2010/cvpr2010.pdf)

De-convolution could be implemented internally via `tf.nn.conv2dBackpropInput` or `tf.nn.conv3dBackpropInput`, also you need to implement analogue of `convOutputLength` function.

Also, if needed, you can take a look at Keras documentation for [`Conv2DTranspose`](https://keras.io/api/layers/convolution_layers/convolution2d_transpose/), and [`Conv3DTranspose`](https://keras.io/api/layers/convolution_layers/convolution3d_transpose/). 

**NOTE**: for the moment, there is no need to add support for "data format" (i.e., channels last vs. channels first) in your PR; you can assume the channels are always in the last dimension.

[1]: https://github.com/JetBrains/KotlinDL/blob/master/api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/convolutional/Conv2D.kt

[2]: https://github.com/JetBrains/KotlinDL/blob/master/api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/inference/keras/ModelSaver.kt

[3]: https://github.com/JetBrains/KotlinDL/blob/master/api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/inference/keras/ModelLoader.kt

[4]:
https://github.com/JetBrains/KotlinDL/blob/master/api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/convolutional/Conv3D.kt

P.S. If you want to take this ticket, please leave the comment below
P.P.S Read the [Contributing Guidelines](https://github.com/JetBrains/KotlinDL/blob/master/CONTRIBUTING.md).

juliabeliaeva:
This PR adds implementations for `Conv1DTranspose`, `Conv2DTranspose` and `Conv3DTranspose` layers.

Caveats:
1. Dilation values greater than 1 are not supported in tensorflow on cpu: https://github.com/tensorflow/tensorflow/issues/28264. Therefore, I have not added any tests for dilations as they do not work.
2. Output padding is not supported in `org.tensorflow.op.NnOps#conv3dBackpropInput`. `Conv3DTranspose` still has the parameter, it just does not do anything. I'm not sure if this parameter is needed or if it is, where to add a warning about its usage.

Fixes #124

### DIFF
--- a/api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/convolutional/Conv1D.kt
+++ b/api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/convolutional/Conv1D.kt
@@ -12,6 +12,7 @@ import org.jetbrains.kotlinx.dl.api.core.initializer.Initializer
 import org.jetbrains.kotlinx.dl.api.core.layer.requireArraySize
 import org.jetbrains.kotlinx.dl.api.core.layer.toLongList
 import org.jetbrains.kotlinx.dl.api.core.regularizer.Regularizer
+import org.jetbrains.kotlinx.dl.api.core.shape.TensorShape
 import org.jetbrains.kotlinx.dl.api.core.shape.convOutputLength
 import org.jetbrains.kotlinx.dl.api.core.util.convBiasVarName
 import org.jetbrains.kotlinx.dl.api.core.util.convKernelVarName
@@ -20,8 +21,6 @@ import org.tensorflow.Shape
 import org.tensorflow.op.Ops
 import org.tensorflow.op.core.Squeeze
 import org.tensorflow.op.nn.Conv2d
-
-private const val EXTRA_DIM = 1L
 
 /**
  * 1D convolution layer (e.g. convolution over audio data).
@@ -79,10 +78,6 @@ public class Conv1D(
 
     override val kernelSize: IntArray = intArrayOf(kernelLength)
 
-    /** Axis of height for which the extra dimension is added (unsqueezed) before actual
-     * convolution operation and the output from actual implementation are squeezed. */
-    private val squeezeAxis = Squeeze.axis(listOf(EXTRA_DIM))
-
     override fun kernelVarName(name: String): String = convKernelVarName(name, dim = 1)
 
     override fun biasVarName(name: String): String = convBiasVarName(name, dim = 1)
@@ -91,15 +86,13 @@ public class Conv1D(
         tf: Ops,
         input: Operand<Float>
     ): Operand<Float> {
-        val expandedInput = tf.expandDims(input, tf.constant(EXTRA_DIM))
-        val expandedKernel = tf.expandDims(kernel.variable, tf.constant(EXTRA_DIM - 1))
-        val expandedStrides = intArrayOf(strides[0], 1, strides[1], strides[2])
-        val expandedDilations = intArrayOf(dilations[0], 1, dilations[1], dilations[2])
-        val options = Conv2d.dilations(expandedDilations.toLongList()).dataFormat("NHWC")
-        val result = tf.nn.conv2d(
-            expandedInput, expandedKernel, expandedStrides.toLongList(), padding.paddingName, options
-        )
-        return tf.squeeze(result, squeezeAxis)
+        return tf.withExpandedDimensions(input) { expandedInput ->
+            val options = Conv2d.dilations(expand(dilations).toLongList()).dataFormat("NHWC")
+            return@withExpandedDimensions tf.nn.conv2d(
+                expandedInput, tf.expandKernel(kernel.variable), expand(strides).toLongList(),
+                padding.paddingName, options
+            )
+        }
     }
 
     protected override fun defineOutputShape(inputShape: Shape): Shape {
@@ -121,5 +114,52 @@ public class Conv1D(
         return "Conv1D(name = $name, isTrainable=$isTrainable, filters=$filters, kernelSize=$kernelSize, strides=${strides.contentToString()}, dilations=${dilations.contentToString()}, activation=$activation, kernelInitializer=$kernelInitializer, biasInitializer=$biasInitializer, kernelRegularizer=$kernelRegularizer, biasRegularizer=$biasRegularizer, activityRegularizer=$activityRegularizer, padding=$padding, useBias=$useBias, hasActivation=$hasActivation, kernelShapeArray=${kernelShapeArray?.contentToString()}, biasShapeArray=${biasShapeArray?.contentToString()})"
     }
 
+    internal companion object {
+        internal const val EXTRA_DIM = 1
 
+        /** Axis of height for which the extra dimension is added (unsqueezed) before actual
+         * convolution operation and the output from actual implementation are squeezed. */
+        private val squeezeAxis = Squeeze.axis(listOf(EXTRA_DIM.toLong()))
+
+        internal fun expandKernel(kernel: IntArray): IntArray {
+            return kernel.withAdded(EXTRA_DIM - 1, 1)
+        }
+
+        internal fun Ops.expandKernel(kernel: Operand<Float>): Operand<Float> {
+            return expandDims(kernel, constant(EXTRA_DIM - 1))
+        }
+
+        internal fun expand(tensorShape: TensorShape): TensorShape {
+            return TensorShape(tensorShape.dims().withAdded(EXTRA_DIM, 1))
+        }
+
+        internal fun expand(array: IntArray): IntArray {
+            return array.withAdded(EXTRA_DIM, 1)
+        }
+
+        /**
+         * Adds an extra dimension to the input, performs the provided operation
+         * and squeezes the result by removing the dimension added previously.
+         * This allows to perform 2D operations on 1D inputs.
+         */
+        internal fun Ops.withExpandedDimensions(input: Operand<Float>,
+                                                operation: (Operand<Float>) -> Operand<Float>
+        ): Operand<Float> {
+            val expandedInput = expandDims(input, constant(EXTRA_DIM))
+            val expandedOutput = operation(expandedInput)
+            return squeeze(expandedOutput, squeezeAxis)
+        }
+
+        internal fun LongArray.withAdded(position: Int, element: Long): LongArray {
+            return toMutableList().apply { add(position, element) }.toLongArray()
+        }
+
+        internal fun IntArray.withAdded(position: Int, element: Int): IntArray {
+            return toMutableList().apply { add(position, element) }.toIntArray()
+        }
+
+        internal fun IntArray.withAdded(position: Int, elements: List<Int>): IntArray {
+            return toMutableList().apply { addAll(position, elements) }.toIntArray()
+        }
+    }
 }
--- a/api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/convolutional/Conv1D.kt
+++ b/api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/convolutional/Conv1D.kt
@@ -38,7 +38,7 @@ private const val EXTRA_DIM = 1L
  * ```
  *
  * @property [filters] The dimensionality of the output space (i.e. the number of filters in the convolution).
- * @property [kernelSize] Long number, specifying the width of the 1D convolution window.
+ * @property [kernelLength] Long number, specifying the width of the 1D convolution window.
  * @property [strides] Three numbers specifying the strides of the pooling operation for each dimension of input tensor.
  * NOTE: Specifying stride value != 1 is incompatible with specifying `dilation` value != 1.
  * @property [dilations] Three numbers specifying the dilation rate to use for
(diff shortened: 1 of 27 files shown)


zaleslaw:
I suggest to be an honest library for a user and remove unsupported parameters by adding to the KDoc the NOTE about this removal (with a link to the known bug).

zaleslaw:
I made an integration example with auto-encoder based on the following article https://www.machinecurve.com/index.php/2019/12/10/conv2dtranspose-using-2d-transposed-convolutions-with-keras/

The Colab is available [here](https://colab.research.google.com/drive/1fyayi9rUa3kXkmrKXcmwRZM1OHIU8SFW?usp=sharing), you could a copy of it and download the JSON config and h5 weights put it to examples/resources, and run the following example

``` 
fun loadAutoencoderModelWithWeightsAndEvaluate() {
    val (_, test) = fashionMnist()

    val jsonConfigFile = getAutoencoderJSONConfigFile()
    val model = Sequential.loadModelConfiguration(jsonConfigFile)

    model.use {
        for(layer in it.layers) {
            layer.isTrainable = false
        }
        it.compile(
            optimizer = Adam(),
            loss = Losses.MAE,
            metric = Metrics.ACCURACY
        )

        it.logSummary()

        val hdfFile = getAutoencoderWeightsFile()
        it.loadWeights(hdfFile)

        val result = it.predict(test.x[0])
    }
}

/** */
fun main(): Unit = loadAutoencoderModelWithWeightsAndEvaluate()

/** Returns JSON file with model configuration, saved from Keras 2.x. */
fun getAutoencoderJSONConfigFile(): File {
    val pathToConfig = "models/mnist/autoencoder/modelConfig.json"
    val realPathToConfig = OnHeapDataset::class.java.classLoader.getResource(pathToConfig).path.toString()

    return File(realPathToConfig)
}

/** Returns .h5 file with model weights, saved from Keras 2.x. */
fun getAutoencoderWeightsFile(): HdfFile {
    val pathToWeights = "models/mnist/autoencoder/weights.h5"
    val realPathToWeights = OnHeapDataset::class.java.classLoader.getResource(pathToWeights).path.toString()
    val file = File(realPathToWeights)
    return HdfFile(file)
}
```

It fails with the following error message
```
Exception in thread "main" java.lang.IllegalStateException: Attempting to use uninitialized value Exception in thread "main" java.lang.IllegalStateException: Attempting to use uninitialized value conv2d_transpose_6_conv2d_transpose_kernel
	 [[{{node Conv2dBackpropInput}}]]
```

Looks like something was missed during the loading of the weights or initialization of the variables.

To reproduce the bug, download the following archive (json + h5)
[weights.zip](https://github.com/JetBrains/KotlinDL/files/7779266/weights.zip)

juliabeliaeva:
@zaleslaw thank you, I forgot about loading weights from h5 files completely. I'll implement it.

juliabeliaeva:
> But it could be a problem during the model convertation Keras->KotlinDL and KotlinDL->Keras and in the case
Keras->KotlinDL we could just skip values and the case KotlinDL->Keras write some default values from Keras layers contructors

That was my concern: if someone loads a model from keras and saves it back, they will loose their parameters in the process.

zaleslaw:
@juliabeliaeva Is this the final version of this PR or is it required any PR should be merged before?

juliabeliaeva:
> @juliabeliaeva Is this the final version of this PR or is it required any PR should be merged before?

It's the final version, can be merged right now.

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/convolutional/Conv1D.kt
```diff
@@ -57,42 +56,27 @@ private const val EXTRA_DIM = 1L
  * @since 0.3
  */
 public class Conv1D(
-    public val filters: Int = 32,
-    public val kernelSize: Int = 3,
```
zaleslaw, line 61: Looks like  kernel size is a stable term in Deep Learning for CNN

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/convolutional/Conv1D.kt
```diff
@@ -116,18 +101,67 @@ public class Conv1D(
...
+        return Shape.make(batchSize, cols, filters.toLong())
     }
 
     override fun toString(): String =
         "Conv1D(filters=$filters, kernelSize=$kernelSize, strides=${strides.contentToString()}, " +
                 "dilation=${dilations.contentToString()}, activation=$activation, kernelInitializer=$kernelInitializer, " +
                 "biasInitializer=$biasInitializer, kernelShape=${kernel.shape}, biasShape=${bias?.shape}, padding=$padding, " +
                 "biasRegularizer=$biasRegularizer, kernelRegularizer=$kernelRegularizer, activityRegularizer=$activityRegularizer)"
+
+    internal companion object {
```
zaleslaw, line 119: Should these functions be companion functions or could be just private functions in the Conv1D?

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/convolutional/Conv1D.kt
```diff
@@ -116,18 +101,67 @@ public class Conv1D(
...
+
+        /** Axis of height for which the extra dimension is added (unsqueezed) before actual
+         * convolution operation and the output from actual implementation are squeezed. */
+        private val squeezeAxis = Squeeze.axis(listOf(EXTRA_DIM.toLong()))
+
+        internal fun expandKernel(kernel: IntArray): IntArray {
+            return kernel.withAdded(EXTRA_DIM - 1, 1)
+        }
+
+        internal fun Ops.expandKernel(kernel: Operand<Float>): Operand<Float> {
```
zaleslaw, line 130: Is it Conv1D specific functions, could they be a part of our local helper framework over the TensorFlow Java API

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/convolutional/Conv1DTranspose.kt
```diff
@@ -0,0 +1,103 @@
...
+ * @property [biasRegularizer] regularizer for the bias
+ * @property [activityRegularizer] regularizer function applied to the output of the layer
+ * @property [padding] type of padding to use
+ * @property [outputPadding] the amount of explicit padding to use (six numbers: two for each dimension).
+ * @property [useBias] a flag that specifies if the bias should be used
+ * @param [name] custom layer name
+ */
+public class Conv1DTranspose(
+    public override val filters: Int = 3,
+    public val kernelLength: Int = 3,
```
zaleslaw, line 59: Kernel Size is better

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/convolutional/Conv1DTranspose.kt
```diff
@@ -0,0 +1,103 @@
...
+        if (outputPadding != null) requireArraySize(outputPadding, 2 * (dimensions + 2), "outputPadding")
+        isTrainable = false
+    }
+
+    override val kernelSize: IntArray = intArrayOf(kernelLength)
+
+    override fun convImplementation(tf: Ops, input: Operand<Float>): Operand<Float> {
+        return tf.withExpandedDimensions(input) { expandedInput ->
+            val expandedOutputPadding = outputPadding?.withAdded(EXTRA_DIM * 2, listOf(0, 0))
+            return@withExpandedDimensions tf.nn.conv2dBackpropInput(
```
zaleslaw, line 86: very complex function call, could we extract some logic to the separate variables and leave some comments for the future (why, for example, we did` tf.shapeWithDynamicBatchSize(outputShape.expand(), input)` for the specific parameter and so on) At this moment this knowledge is hidden in the issue comments.
I suggest doing this, because it's enough far from the initial Keras code

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/convolutional/Conv2DTranspose.kt
```diff
@@ -0,0 +1,126 @@
...
+                outputPadding?.withStandardPadding(
+                    padding,
+                    kernelSize,
+                    dilations
+                )
+            )
+        )
+    }
+
+    internal companion object {
```
zaleslaw, line 97: Methods of this companion are used in the both, Conv1DTranspose and Conv2DTranspose, probably need we need a separate place with top-level functions a la Util class.

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/convolutional/Conv3DTranspose.kt
```diff
@@ -0,0 +1,88 @@
...
+ * This layer expects input data of size `(N, D, H, W, C)` where
+ * ```
+ * N - batch size
+ * D - depth
+ * H - height
+ * W - width
+ * C - number of channels
+ * ```
+ *
+ * Note: providing explicit output padding is currently not supported.
```
zaleslaw, line 35: Should we throw an exception on that or add a contract to `init `section?

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/convolutional/ConvTranspose.kt
```diff
@@ -0,0 +1,92 @@
...
+                "biasRegularizer=$biasRegularizer, " +
+                "kernelRegularizer=$kernelRegularizer, " +
+                "activityRegularizer=$activityRegularizer, " +
+                "padding=$padding, " +
+                "outputPadding=${outputPadding?.contentToString()} " +
+                ")"
+    }
+
+    internal companion object {
+        /**
```
zaleslaw, line 77: Could we move here other companion functions mentioned earlier?

examples/src/main/kotlin/examples/cnn/fsdd/SoundNet.kt
```diff
@@ -41,7 +41,7 @@ internal fun soundBlock(filters: Int, kernelSize: Int, poolStride: Int): Array<L
     arrayOf(
         Conv1D(
             filters = filters,
-            kernelSize = kernelSize,
+            kernelLength = kernelSize,
```
zaleslaw, line 44: kernelSize!

zaleslaw (changes requested):
Please have a look at the left comments and add the general example related to the autoencoder based on the Colab shared earlier in the issue comments.
No big changes, just a few minor things for future developers.

At this moment #328 is merged to the master, you could merge in this PR and finish the Autoencoder example

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/convolutional/Conv1D.kt
juliabeliaeva: We can't use `kernelSize` here since `kernelSize` is an array and we want to allow passing a single integer here.

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/convolutional/Conv1D.kt
juliabeliaeva: These functions are used in both `Conv1D` and `Conv1DTransoise` since 1D convolutions are implemented with 2D convolutions.

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/convolutional/Conv1D.kt
juliabeliaeva: These functions are `Conv1D`/`Conv1DTranspose`, I don't think they are needed outside of these operations.

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/convolutional/Conv3DTranspose.kt
juliabeliaeva: There is no `outputPadding` parameter, so there is no situation where we could throw an exception. The note is here since keras does have this parameter (see https://keras.io/api/layers/convolution_layers/convolution3d_transpose/).

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/convolutional/Conv1DTranspose.kt
juliabeliaeva: > why, for example, we did tf.shapeWithDynamicBatchSize(outputShape.expand(), input)

`shapeWithDynamicBatchSize` has a javadoc with explanation and links to the relevant issues, I'll add some comments to this method as well.

juliabeliaeva:
Pushed a new version of this PR.
1. Rebased on master, resolved conflicts.
2. Rewritten `Conv1DTranspose` implementation to be more clear.
3. Moved helper methods.
4. Discovered that my implementation computes the output size a bit differently than keras, so I rewritten it to match (see the last commit in the branch).

PR #277, opened 2021-11-12
Introduced Parameter data class to encapsulate all information about single parameter of layer

knok16:

PR #318, opened 2021-12-28
Simplify loading layer weights support for new layers

juliabeliaeva:
The goal of this PR is to make adding support for the new layers to the `WeightLoader` a little bit easier and help with finishing https://github.com/JetBrains/KotlinDL/pull/315.
The code for specific layer classes is extracted into `WeightMappings` file. For each layer there are two methods: `get<LAYER_NAME>Variables` and `get<LAYER_NAME>VariablePathTemplates` which provide information about layer variables and how to read them. In the future, after https://github.com/JetBrains/KotlinDL/pull/277 is merged, these methods will use `KVariable` and become even simpler.
This PR also adds warnings when a layer in the model is not yet supported.


zaleslaw:
@juliabeliaeva do you need this to be merged before #315?

juliabeliaeva:
> @juliabeliaeva do you need this to be merged before #315?

@zaleslaw yes.

PR #320, opened 2021-12-30
Added Support For ByteArray Image Loading

AlexandreBrown:
This is an attempt to add support for loading byte array in the image loading preprocessing DSL.  
  
This PR is not ready yet, there are still big design changes needed but it can start a discussion.  
  
Relates to issue https://github.com/JetBrains/KotlinDL/issues/321

### DIFF
--- a/api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/inference/imagerecognition/ImageRecognitionModel.kt
+++ b/api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/inference/imagerecognition/ImageRecognitionModel.kt
@@ -16,6 +16,8 @@ import org.jetbrains.kotlinx.dl.dataset.preprocessor.image.InterpolationType
 import org.jetbrains.kotlinx.dl.dataset.preprocessor.image.convert
 import org.jetbrains.kotlinx.dl.dataset.preprocessor.image.resize
 import java.io.File
+import org.jetbrains.kotlinx.dl.dataset.preprocessor.image.ImageFormatType
+import org.jetbrains.kotlinx.dl.dataset.preprocessor.image.loading.FileLoadingStrategy
 
 /**
  * The light-weight API for solving Image Recognition task with one of the Model Hub models trained on ImageNet dataset.
@@ -74,7 +76,10 @@ public class ImageRecognitionModel(
 
         val preprocessing: Preprocessing = preprocess {
             load {
-                pathToData = imageFile
+                loadingStrategy = FileLoadingStrategy(
+                    imageFormat = ImageFormatType.valueOf(imageFile.extension),
+                    imageFile = imageFile
+                )
                 imageShape = ImageShape(null, null, 3)
             }
             transformImage {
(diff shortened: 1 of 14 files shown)


api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/inference/imagerecognition/ImageRecognitionModel.kt
```diff
@@ -74,7 +76,10 @@ public class ImageRecognitionModel(
 
         val preprocessing: Preprocessing = preprocess {
             load {
-                pathToData = imageFile
+                loadingStrategy = FileLoadingStrategy(
+                    imageFormat = ImageFormatType.valueOf(imageFile.extension),
+                    imageFile = imageFile
+                )
```
AlexandreBrown, line 82: Here is an example of how it would be used so far

AlexandreBrown: For an image in bytes we would have  
```kotlin
load {
                loadingStrategy = BytesLoadingStrategy(
                    imageFormat = ImageFormatType.PNG,
                    imageBytes = myImageBytes
                )
                imageShape = ImageShape(null, null, 3)
            }
```

juliabeliaeva:
Hi, thank you for the PR. The functionality of preprocessing images already loaded into the memory is indeed very useful. But in my opinion, it should be implemented differently. In my mind, the process of loading images (from files or other sources) should in some form be a part of the dataset, not preprocessing, as it looks like the knowledge about the source of the images is intrinsic to the dataset implementation. Take for example `OnFlyImageDataset`. It already heavily depends on the preprocessing loading stage as it keeps files with images in the field and uses them to load data on the fly (see [`OnFlyImageDataset#init`](https://github.com/JetBrains/KotlinDL/blob/ebfe99c0dfbc424e9e7795c7d3e6528ad5d19cd8/dataset/src/main/kotlin/org/jetbrains/kotlinx/dl/dataset/OnFlyImageDataset.kt#L30)). In principle, this code could be changed to load images from elsewhere, but having images as byte arrays in memory would defy the purpose of this class. The other `Dataset` implementation, `OnHeapDataset` can actually be made to work with byte arrays. To do so, one can implement a method similar to  [`OnHeapDataset#create`](https://github.com/JetBrains/KotlinDL/blob/ebfe99c0dfbc424e9e7795c7d3e6528ad5d19cd8/dataset/src/main/kotlin/org/jetbrains/kotlinx/dl/dataset/OnHeapDataset.kt#L208), but without accessing loading stage of the preprocessing function (and probably asserting it to be not set) and just accepting images as a parameter.
If loading stage is moved out preprocessing, then preprocessing would become just a filter able to work on any images, either from a dataset or loaded manually from any sources.
What do you think about this approach?

AlexandreBrown:
Hello @juliabeliaeva , I very much prefer your approach and I agree with the fact that the dataset has the responsibility to load the data therefore it would make more sense to have it implemented in the dataset (and it might simplify the implementation).  
  
I have not looked into `OnHeadDataset` too much, in fact this was the last thing to check for my proposal but I will scrap the proposal and follow your suggestion to make the changes mostly in the `Dataset` objects instead.
  
Just a remark, I think it's important to think about an approach that will be easily grasped whether you want to preprocess for training or for inference. My initial thought was to implement it in the `Loading` preprocessing because the examples propose to use only the preprocessing dsl when we just want to preprocess an image for inference for instance.   
Your idea of having it in the Dataset object makes perfect sense but let's not forget to update the examples and make it clear that whether you want to preprocess for training or inference, you gotta create a `Dataset` object.

PR #326, opened 2022-01-13
Remove duplication between methods loading weights and cleanup code

### LINKED ISSUE #325: Code duplication in `WeightLoader`
`WeightLoader` has a number of methods to load all or a subset of layers in variety of ways, but essentially all of them are performing the same task:
1. Check that model can load weights.
2. Load weights for a subset of layers, and initialise weights for other layers.
3. Mark model as initialised.

Refactoring is needed to unify these methods.

juliabeliaeva:
Fixes #325

### DIFF
--- a/api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/inference/keras/WeightLoader.kt
+++ b/api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/inference/keras/WeightLoader.kt
@@ -27,9 +27,9 @@ import org.jetbrains.kotlinx.dl.api.inference.keras.WeightMappings.getLayerVaria
 public fun GraphTrainableModel.loadWeights(
     hdfFile: HdfFile
 ) {
-    check(this.isModelCompiled) { "The model is not compiled yet. Compile the model to use this method." }
+    check(isModelCompiled) { "The model is not compiled yet. Compile the model to use this method." }
     check(!isModelInitialized) { "Model is initialized already!" }
-    this.logger.info { "Starting weights loading.." }
+    logger.info { "Starting weights loading.." }
 
     when {
         hdfFile.attributes.containsKey("layer_names") -> loadWeightsFromHdf5Group(hdfFile, this, null)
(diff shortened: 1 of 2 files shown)


juliabeliaeva:
Force-pushed to fix a test (error wording changed).

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/inference/keras/WeightLoader.kt
```diff
@@ -183,108 +86,49 @@ public fun GraphTrainableModel.loadWeightsByPathTemplates(
...
- * NOTE: Weights for another layers will not be loaded (should be initialized manually).
+ * NOTE: Weights for trainable layers will not be loaded and will be initialized via default initializers.
  *
  * @param [hdfFile] File in hdf5 file format containing weights of Sequential model.
- * @param [layerList] List of layers to load weights. Weights for other layers will be initialized by initializer later.
  * @param [kernelDataPathTemplate] Template path to kernel weights of the specific layer.
  * @param [biasDataPathTemplate] Template path to bias weights of the specific layer.
  */
-public fun GraphTrainableModel.loadWeightsByPathTemplates(
+public fun Functional.loadWeightsForFrozenLayersByPathTemplates(
```
zaleslaw, line 100: @juliabeliaeva could you please explain excluding the Sequential from the ability to have this function, I suppose it could the common functionality (agree, that there is no examples in the codebase)

juliabeliaeva: It was like this in `master`, I did not change it. See https://github.com/JetBrains/KotlinDL/blob/11d4eb175323b61d2e709d4133cbbde49c033746/api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/inference/keras/WeightLoader.kt#L278

juliabeliaeva: Changed the type to `GraphTrainableModel`.

zaleslaw: Great!

PR #332, opened 2022-02-01
Added support for multiple Callbacks in fit(), evaluate(), predict() instead of compile() method

zaleslaw:


zaleslaw:
@rychuelektryk could you please have a look at the proposed PR with the new example for the multiple callbacks that could be passed via fit/evaluate/predict method. Does it solve your problem? https://github.com/JetBrains/KotlinDL/pull/332/files#diff-c6505be13ba6fb2e551cb414c39c471e4ad77265e272e461584ecfd8f3b6d640

rychuelektryk (commented):
Looks ok

zaleslaw:
Thanks, @rychuelektryk for the quick response

PR #345, opened 2022-03-25
Introduce ParametrizedLayer and TrainableLayer interfaces

### LINKED ISSUE #217: [API] [Layers] ParametrizedLayer/TrainableLayer interfaces to represent presence of parameters and trainability of the layer
Hello, what about moving some aspects of Layer logic into separate interfaces, to offload Layer class?

For example, introduce ParametrizedLayer/TrainableLayer interfaces to encapsulate `isTranable` / `paramCount` logic and will be implemented only by layers that actually have variables/can be trained

```
public interface ParametrizedLayer {
    /**
     * Layer's variables
     */
    public val variables: List<Variable>

    /**
     * Returns amount of parameters
     */
    public val ParametrizedLayer.paramCount: Int
        get() = variables.sumOf { it.shape.numElements() }.toInt()
}

public interface TrainableLayer : ParametrizedLayer {
    /**
     * True, if layer's weights could be changed during training.
     * If false, layer's weights are frozen and could not be changed during the training.
     */
    public var isTrainable: Boolean
}
```

Benefits:
- Layer class itself will become thinner and hence easier to grasp
- Untrainable by nature (for example `Flatten`,`Input`,`ActivationLayer`) layers will not even have `isTrainable` flag and `weight` field
- Main idea here is that the code `Flatten().isTrainable = true` logically doesn't make sense, and with `ParametrizedLayer` it will not compile
- Instead of marking layers with `NoGradient` it will be possible just not to implement `Trainable` which will work as whitelisting for training (instead of current blacklisting) which is more explicit.
- It will be possible to remove KGraph parameter from Layer::build method
- It will be possible to move `weight` setter/getter to the model (example https://github.com/knok16/KotlinDL/commit/4dc0286a4773eb3e5970869b7cff8c918ac8cb6c)

Reference https://github.com/knok16/KotlinDL/commit/f5244b6cd850e9e7026fb71a13174cc13e9c067e

juliabeliaeva:
This PR adds the following changes:
1. Introduce `ParametrizedLayer` and `TrainableLayer` interfaces and utility methods to work with variables and weights.
2. Move variable information to the layers, instead of keeping it in `KGraph`.

Fixes #217

### DIFF
--- a/api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/GraphTrainableModel.kt
+++ b/api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/GraphTrainableModel.kt
@@ -114,8 +114,8 @@ public abstract class GraphTrainableModel(vararg layers: Layer) : TrainableModel
         session = Session(kGraph.tfGraph)
     }
 
-    override fun variables(): List<Variable<Float>> = layers.tfVariables()
-    override fun frozenVariables(): List<Variable<Float>> = layers.frozenTfVariables()
+    override fun variables(): List<Variable<Float>> = layers.variables().map { it.variable }
+    override fun frozenVariables(): List<Variable<Float>> = layers.frozenVariables().map { it.variable }
 
     /** Helper method for preprocessing layer names and layer validation. */
     internal companion object {
@@ -173,7 +173,7 @@ public abstract class GraphTrainableModel(vararg layers: Layer) : TrainableModel
 
         yPredOp = forward(xOp, inputLayer)
         lossOp = buildLossFunction(loss)
-        targets = optimizer.prepareTargets(kGraph, layers.trainableTfVariables(), tf, lossOp)
+        targets = optimizer.prepareTargets(kGraph, layers.trainableVariables().map { it.variable }, tf, lossOp)
 
         predictionOp = when (loss) {
             is SoftmaxCrossEntropyWithLogits -> tf.withName(OUTPUT_NAME).nn.softmax(yPredOp)
--- a/api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/GraphTrainableModel.kt
+++ b/api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/GraphTrainableModel.kt
@@ -191,7 +191,7 @@ public abstract class GraphTrainableModel(vararg layers: Layer) : TrainableModel
         val basicLoss = loss.apply(tf, yPredOp, yTrueOp, numberOfLossesOp)
         var totalLoss = basicLoss
         // TODO: probably regularization output should be divided on numberOfLossesOp and changed together with loss before averaging
-        layers.variables().forEach { variable ->
+        layers.trainableVariables().forEach { variable ->
             variable.regularizer?.let { regularizer ->
                 totalLoss = tf.math.add(totalLoss, regularizer.apply(tf, variable.variable))
             }
--- a/api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/GraphTrainableModel.kt
+++ b/api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/GraphTrainableModel.kt
@@ -958,3 +958,11 @@ public abstract class GraphTrainableModel(vararg layers: Layer) : TrainableModel
         )
     }
 }
+
+/**
+ * Freezes weights in all layers in this model, so they won't be changed during training.
+ * @see [Layer.freeze]
+ */
+public fun GraphTrainableModel.freeze() {
+    layers.forEach(Layer::freeze)
+}
--- a/api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/GraphTrainableModel.kt
+++ b/api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/GraphTrainableModel.kt
@@ -11,10 +11,11 @@ import org.jetbrains.kotlinx.dl.api.core.callback.Callback
 import org.jetbrains.kotlinx.dl.api.core.exception.RepeatableLayerNameException
 import org.jetbrains.kotlinx.dl.api.core.history.*
 import org.jetbrains.kotlinx.dl.api.core.layer.Layer
-import org.jetbrains.kotlinx.dl.api.core.layer.NoGradients
 import org.jetbrains.kotlinx.dl.api.core.layer.core.ActivationLayer
 import org.jetbrains.kotlinx.dl.api.core.layer.core.Dense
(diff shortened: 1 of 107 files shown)


zaleslaw:
So, I'm planning to make a review for this huge PR next week, if it's ok

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/GraphTrainableModel.kt
```diff
@@ -114,6 +114,9 @@ public abstract class GraphTrainableModel(vararg layers: Layer) : TrainableModel
         session = Session(kGraph.tfGraph)
     }
 
+    override fun variables(): List<Variable<Float>> = layers.tfVariables()
```
zaleslaw, line 117: probably rename to just 'variables'

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/activation/PReLU.kt
```diff
@@ -36,7 +36,7 @@ public class PReLU(
     public val alphaRegularizer: Regularizer? = null,
     public val sharedAxes: IntArray? = null,
     name: String = ""
-) : AbstractActivationLayer(name) {
+) : AbstractActivationLayer(name), ParametrizedLayer, TrainableLayer {
```
zaleslaw, line 39: But TrainableLayer is a sub-type for Parametrized? Could we remove Parametrized layer from the signature?

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/normalization/BatchNorm.kt
```diff
@@ -57,31 +52,27 @@ public class BatchNorm(
     public val movingMeanInitializer: Initializer = Zeros(),
     public val movingVarianceInitializer: Initializer = Ones(),
     name: String = "",
-) : Layer(name), NoGradients {
+) : Layer(name), NoGradients, ParametrizedLayer {
```
zaleslaw, line 55: Could we make NoGradients a subtype of Parametrized but notTrainable?

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/inference/keras/loaders/TFModelHub.kt
```diff
@@ -178,9 +180,7 @@ public class TFModelHub(cacheDirectory: File) : ModelHub(cacheDirectory) {
     }
 
     private fun freezeAllLayers(model: GraphTrainableModel): GraphTrainableModel {
-        for (layer in model.layers) {
-            layer.isTrainable = false
-        }
+        model.layers.forEach(Layer::freeze)
```
zaleslaw, line 183: looks like we could add some sugar here with model.freeze()

api/src/test/kotlin/org/jetbrains/kotlinx/dl/api/core/integration/SequentialInferenceTest.kt
```diff
@@ -370,10 +374,7 @@ class SequentialInferenceTest {
         val model = Sequential.loadDefaultModelConfiguration(tempDir.toFile())
 
         model.use {
-            for (layer in it.layers) {
-                if (layer::class == Conv2D::class)
-                    layer.isTrainable = false
-            }
+            it.layers.filterIsInstance<Conv2D>().forEach(Layer::freeze)
```
zaleslaw, line 377: filterIsInstanse with generic looks worse than we removed a piece of code

How is it implemented internally? Does it iterate two times over the list of layers?

zaleslaw (changes requested):
I've explored this huge PR locally and I've added some minor comments, please fix them or answer

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/normalization/BatchNorm.kt
juliabeliaeva: Actually, no. There is, for example, `Concatenate` layer which is not `Parametrized` but implements `NoGradients`.

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/ParametrizedLayer.kt
```diff
@@ -0,0 +1,81 @@
...
+ * Returns a list of trainable variables used in the layers.
+ */
+internal fun List<Layer>.trainableVariables(): List<KVariable> {
+    return filterIsInstance<TrainableLayer>().filter { it.isTrainable }.flatMap { it.variables }
+}
+
+/**
+ * Returns a list of trainable variables used in the layers.
+ */
+internal fun List<Layer>.trainableTfVariables(): List<Variable<Float>> = trainableVariables().map { it.variable }
```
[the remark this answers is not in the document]
juliabeliaeva: The problem is that we have our `KVariable` class and also `Variable` from tensorflow. So here `trainableVariables()` returns `KVariable` and `trainableTfVariables()` -- `Variable`. If we rename the latter method, we should rename the former. `trainableKVariable()` would be ok? Same goes for the other pairs of methods.

api/src/test/kotlin/org/jetbrains/kotlinx/dl/api/core/integration/SequentialInferenceTest.kt
juliabeliaeva: Yes, it does iterate two times for each `Conv2D` layer. Imao, the model probably does not have that many convolutional layers when it could become a problem.

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/ParametrizedLayer.kt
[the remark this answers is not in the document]
juliabeliaeva: Actually, perhaps, functions returning tensorflow variables aren't needed at all. There aren't that many usages for them.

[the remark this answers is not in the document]
juliabeliaeva: Removed the "tf"-functions.

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/activation/PReLU.kt
juliabeliaeva: Done.

zaleslaw:
I suppose it could be merged as well the merge conflict will be resolved @juliabeliaeva 
Thanks for changes in latest commits

PR #352, opened 2022-04-13
Simplify working with optimizer variables

juliabeliaeva:
This PR deals with optimizer variables created by the `createSlot` function. Instead of creating variables in the `Optimizer#createSlots`, putting them to the special map only to retrieve in the `applyGradients`, which is called directly after, variables are created immediately before usage.


zaleslaw:
It's an interesting proposal, but I need a time to estimate the power of this refactoring, give me a couple of days here to trace all the code routes in refactored optimizers

juliabeliaeva:
@zaleslaw I don't think this refactoring should be included into the 0.4 so there is really no rush with the review.

zaleslaw:
If you suppose to postpone it, I suggest marking it with the 0.5 milestone

PR #355, opened 2022-04-18
Fix variables initialization

### LINKED ISSUE #4: The initializer name schema is violated by Zero and Ones ops
If you will load a graph from Keras in hdf5 format it could keep Zeros or Ones as a default initializer for biases.
But the standard schema of loading weights via initializers names doesn't work well in this way. Should be used another initializer (maybe special) for handling this issue.

Name: Init_conv2d_25_conv2d_bias/Zero; Type: Const; Out #tensors:  1
Name: Init_conv2d_25_conv2d_bias/Fill; Type: Fill; Out #tensors:  1

juliabeliaeva:
This PR fixes #4 by storing the initializer `Operand` in `KVariable` and accessing it directly instead of trying to find it by name. This PR is based on the PR #345.


zaleslaw:
I'll take it for review

zaleslaw:
Sorry, @juliabeliaeva is it ready for review?

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/inference/keras/WeightLoader.kt
```diff
@@ -255,7 +252,7 @@ private fun fillLayerVariablesFromKeras(layerName: String,
         require(variable.shape.toIntArray().contentEquals(dims)) {
             "${variable.name} variable shape in loaded data is ${dims.contentToString()}. Should be ${variable.shape.toIntArray().contentToString()}"
         }
-        model.fillVariable(variable.name, it.data)
```
zaleslaw, line 258: I'm in doubts regarding changing model.fillVariable to variable.fill by model. Session is related to model and initialized variables are too belong to model. Looks like changing of the sense and accents.

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/KVariable.kt
```diff
@@ -19,17 +22,52 @@ import org.tensorflow.op.core.Variable
...
     val regularizer: Regularizer?
 )
 
+/**
+ * A class that keeps information which allows to initialize [KVariable].
+ *
+ * @property [assign] assign operation for the variable
+ * @property [initialValue] value generated by the default initializer
+ */
+public data class InitializerOperation(val assign: Assign<Float>, val initialValue: Operand<Float>) {
```
zaleslaw, line 42: Will it support another type of assign operands, like assignandadd or something else?

juliabeliaeva: Are `AssignAdd` operations usable for variable initialization? Anyway, I can change the type to `Operand<Float>` (same as the initalizer type in `KVariable` before the refactoring).

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/inference/keras/WeightLoader.kt
juliabeliaeva: Moved `init` and `fill` methods to the model.

zaleslaw:
I suppose it could be merged

Issue #285, opened 2021-11-19
Type cast error when inferencing onnx model


gasabr:
I'm trying to inference ONNX model created from lightgbm model via Kotlin DL and in every method (tried Raw ones too) i'm getting `class [J cannot be cast to class [[F ([J and [[F are in module java.base of loader 'bootstrap')` or in RawMethods `SequenceInfo cannot be cast to class ai.onnxruntime.TensorInfo`

Env:
```
    implementation("org.jetbrains.kotlinx:kotlin-deeplearning-api:0.3.0")
    implementation("org.jetbrains.kotlinx:kotlin-deeplearning-onnx:0.3.0-alpha-3")
```

Code
```
package co.`fun`

import kotlin.test.*
import org.jetbrains.kotlinx.dl.api.inference.onnx.OnnxInferenceModel
import kotlin.random.Random

class ApplicationTest {
    @Test
    fun testRoot() 
[twoTierModel.txt](https://github.com/JetBrains/KotlinDL/files/7570558/twoTierModel.txt)
{
        val onnxModel = OnnxInferenceModel.load("/tmp/twoTierModel.onnx")
        onnxModel.reshape(27)
        val features = (1..27).map { Random.nextFloat() }.toFloatArray()
        val prediction = onnxModel.predictSoftly(features, "features")
    }
}
```

Error is in the line 124 of the file `OnnxInferenceModel.kt` and it's caused by the attempt to cast List<INT64> to Array<FloatArray>, I'm not sure if the model should always return 3d Tensor or the lib should check the types.

Rename the attachment to `twoTierModel.onnx` to try the test at your machine

zaleslaw:
Please, add the attachment @gasabr and could you try to repeat the experiment with the latest version of the onnx-dependency, here
```
    implementation("org.jetbrains.kotlinx:kotlin-deeplearning-api:0.3.0")
    implementation("org.jetbrains.kotlinx:kotlin-deeplearning-onnx:0.3.0")
```

gasabr:
[twoTierModel.txt](https://github.com/JetBrains/KotlinDL/files/7579067/twoTierModel.txt)

Thanks for the response! I tried 0.3.0 got the same exception, I also have tried to use java onnxruntime and was able to inference the model with following code:

```
        val env = OrtEnvironment.getEnvironment()
        val session = env.createSession("/tmp/twoTierModel.onnx", OrtSession.SessionOptions())
[...]

PR #356, opened 2022-04-18
Add model output type checks for OnnxInferenceModel (#285)

ermolenkodev:
* If model output type not supported exception will be thrown

    * Exception message contains recommendation to use predictRaw() method

    * Add test demonstrating usage of onnx model with unsupported output type

    * Minor changes after apply of auto-format


ermolenkodev:
It's not an actual fix of [#285 ](https://github.com/JetBrains/KotlinDL/issues/285) but temporary solution to facilitate releasing of 0.4. Later we will implement support of ONNX output types such as Sequence and Map

onnx/src/main/kotlin/org/jetbrains/kotlinx/dl/api/inference/onnx/OnnxInferenceModel.kt
```diff
@@ -117,7 +117,32 @@ public open class OnnxInferenceModel : InferenceModel() {
...
     override fun predictSoftly(inputData: FloatArray, predictionTensorName: String): FloatArray {
-        return predictSoftly(inputData)
+        require(::inputShape.isInitialized) { "Reshape functions is missed! Define and set up the reshape function to transform initial data to the model input." }
+
+        val outputTensorName = when {
+            predictionTensorName.isEmpty() -> session.outputNames.first()
+            else -> predictionTensorName
+        }
+
+        require(outputTensorName in session.outputInfo) { "There is no output with name $outputTensorName" }
```
juliabeliaeva, line 127: We could add the available names to the error message.

onnx/src/main/kotlin/org/jetbrains/kotlinx/dl/api/inference/onnx/OnnxInferenceModel.kt
```diff
@@ -127,21 +152,27 @@ public open class OnnxInferenceModel : InferenceModel() {
      * @return Vector that represents the probability distributions of a list of potential outcomes
      */
     public fun predictSoftly(inputData: FloatArray): FloatArray {
-        require(::inputShape.isInitialized) { "Reshape functions is missed! Define and set up the reshape function to transform initial data to the model input." }
+        return predictSoftly(inputData, session.outputNames.first())
+    }
 
+    private fun predictSoftly(inputData: FloatArray, outputTensorIdx: Int): FloatArray {
```
juliabeliaeva, line 158: I think it would be a good idea to unify this method with `predictRaw` (as a separate PR).

juliabeliaeva (changes requested):

onnx/src/main/kotlin/org/jetbrains/kotlinx/dl/api/inference/onnx/OnnxInferenceModel.kt
ermolenkodev: Agree, but the 'predictRaw' method currently has lower performance than other methods. So it requires some research to refactor it properly.

juliabeliaeva (approved):

Issue #89, opened 2021-06-01
Model save/load compatibility with Keras


mkaze:
I was wondering if we are aiming for a two-way compatibility with Keras in terms of saving and loading of the models. In particular, should both of the following conditions satisfy?

- Models trained and saved with Keras could be fully loaded in KotlinDL.
- Models trained and saved with KotlinDL could be fully loaded in Keras.

zaleslaw:
Yes, but the first priority is the loading model from Keras for Transfer Learning purposes. It's an important and popular use case.

Also, we need a good way to save the model for KotlinDL needs (currently, there are a few ways, and one of the exports to Keras-like JSON format); it's the second point here.

It's a good idea to have full export/import with Keras in both directions, hope to return to this task and open a few tickets about that. 

I worry that loading to Keras will not work fully due to issues with weights export (I could not find any library for h5 writing, only for reading) and some missed things in JSON but it could be fixed in the future.

dosier:
I found two potentially useful projects with regards to writing h5
* https://github.com/h5jan/h5jan-core
* https://github.com/saddle/jhdf5

zaleslaw:
P.S Read the [Contributing Guidelines](https://github.com/JetBrains/KotlinDL/blob/master/CONTRIBUTING.md).

PR #360, opened 2022-04-22
Fix layers import and export

juliabeliaeva:
This PR fixes various issues with importing and exporting layers configuration and adds missing weights import for `Conv1D`, `Conv3D` and `PReLU` layers. This is a step towards addressing #89.

### DIFF
--- a/api/src/test/kotlin/org/jetbrains/kotlinx/dl/api/inference/keras/ActivationLayersImportExportTest.kt
+++ b/api/src/test/kotlin/org/jetbrains/kotlinx/dl/api/inference/keras/ActivationLayersImportExportTest.kt
@@ -1,0 +1,92 @@
+/*
+ * Copyright 2022 JetBrains s.r.o. and Kotlin Deep Learning project contributors. All Rights Reserved.
+ * Use of this source code is governed by the Apache 2.0 license that can be found in the LICENSE.txt file.
+ */
+
+package org.jetbrains.kotlinx.dl.api.inference.keras
+
+import org.jetbrains.kotlinx.dl.api.core.Sequential
+import org.jetbrains.kotlinx.dl.api.core.activation.Activations
+import org.jetbrains.kotlinx.dl.api.core.initializer.HeNormal
+import org.jetbrains.kotlinx.dl.api.core.layer.activation.*
+import org.jetbrains.kotlinx.dl.api.core.layer.core.ActivationLayer
+import org.jetbrains.kotlinx.dl.api.core.layer.core.Input
+import org.jetbrains.kotlinx.dl.api.core.regularizer.L2
+import org.junit.jupiter.api.Test
+
+class ActivationLayersImportExportTest {
+    @Test
+    fun activationLayer() {
+        LayerImportExportTest.run(
+            Sequential.of(
+                Input(100),
+                ActivationLayer(name = "test_activation_layer", activation = Activations.Exponential)
+            )
+        )
+    }
+
+    @Test
+    fun elu() {
+        LayerImportExportTest.run(
+            Sequential.of(
+                Input(100),
+                ELU(name = "test_elu", alpha = 0.5f)
+            )
+        )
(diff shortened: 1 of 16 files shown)


api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/util/ContainerUtil.kt
```diff
@@ -20,4 +20,10 @@ public fun <K, V> mapOfNotNull(vararg mapping: Pair<K?, V?>): Map<K, V> {
             }
         }
     }
+}
+
+internal fun Array<IntArray>.deepContentToString(): String {
```
zaleslaw, line 25: Looks like missed thing in the standard library!

api/src/test/kotlin/org/jetbrains/kotlinx/dl/api/inference/keras/ActivationLayersImportExportTest.kt
```diff
@@ -0,0 +1,92 @@
...
+        LayerImportExportTest.run(
+            Sequential.of(
+                Input(100),
+                ELU(name = "test_elu", alpha = 0.5f)
+            )
+        )
+    }
+
+    @Test
+    fun `leaky relu`() {
```
zaleslaw, line 39: Could you please explain or refer to a rule with test namings, somewhere you use short names without `` and somewhere small sentences in ` `.
I suggest using one type of naming according the previous codebase

api/src/test/kotlin/org/jetbrains/kotlinx/dl/api/inference/keras/ConvolutionalLayersImportExportTest.kt
```diff
@@ -162,7 +162,7 @@ class ConvolutionalLayersPersistenceTest {
 
     @Test
     fun separableConvTest() {
```
zaleslaw, line 164: somewhere you use test at the end of test names

api/src/test/kotlin/org/jetbrains/kotlinx/dl/api/inference/keras/MergeLayersImportExportTest.kt
```diff
@@ -0,0 +1,126 @@
...
+
+    @Test
+    fun dot() {
+        LayerImportExportTest.run(
+            functional {
+                layers {
+                    val input = +Input(10)
+                    val dense1 = +Dense(name = "Dense1", outputSize = 5)(input)
+                    val dense2 = +Dense(name = "Dense2", outputSize = 5)(input)
+                    +Dot(name = "test_dot", axis = intArrayOf(1, 1), normalize = true)(dense1, dense2)
```
zaleslaw, line 121: Nice tests!

zaleslaw (commented):
Let's clarify some details regarding test naming

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/util/ContainerUtil.kt
ilya-g: https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/content-deep-to-string.html

api/src/test/kotlin/org/jetbrains/kotlinx/dl/api/inference/keras/ActivationLayersImportExportTest.kt
juliabeliaeva: It is a bit inconsistent, I agree. I looked at the codebase, and it seems that camel-case naming is more prevalent, although there are tests which use spaces (like `org.jetbrains.kotlinx.dl.api.inference.keras.ModelLoaderTest#load initializers from file`). So I changed test function names in this PR so that they use camel-case.

PR #373, opened 2022-05-20
Fix problem with resource resolving on Windows

ermolenkodev:
This fixes the issue which causes OnnxOutputsSupportTestSuite fails on the Windows test agent


zaleslaw:
I suppose it could be merged, fix is correct

juliabeliaeva (approved):

PR #378, opened 2022-05-25
Clamp input values to 0..255 range when converting FloatArray to a BufferedImage

### LINKED ISSUE #377: Clamp input data to 0..255 range in ImageConverter#floatArrayToBufferedImage

juliabeliaeva:
Fixes #377

### DIFF
--- a/dataset/src/main/kotlin/org/jetbrains/kotlinx/dl/dataset/image/ImageConverter.kt
+++ b/dataset/src/main/kotlin/org/jetbrains/kotlinx/dl/dataset/image/ImageConverter.kt
@@ -238,7 +238,7 @@ public object ImageConverter {
         outputShape: ImageShape,
         arrayColorMode: ColorMode,
         isNormalized: Boolean
-    ) : BufferedImage {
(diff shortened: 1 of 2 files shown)


ermolenkodev (commented):
What do you think to move the clamping to another floatArrayToBufferedImage function? like that
```
public fun floatArrayToBufferedImage(
        inputArray: FloatArray,
        outputShape: ImageShape,
        arrayColorMode: ColorMode,
        isNormalized: Boolean
    ): BufferedImage {
        return floatArrayToBufferedImage(inputArray, outputShape, arrayColorMode) {
            if (isNormalized) denormalizeInplace(it, scale = 255f)
            for ((i, value) in dataCopy.withIndex()) {
               if (value < 0) dataCopy[i] = 0f
               if (value > 255) dataCopy[i] = 255f
             }
             it
        }
    }
```
in that case, floatArrayToBufferedImage, which accepts ArrayTransform, will be more flexible in some sense

Also, maybe it's worth adding an explicit comment that values will be clamped.

PR #392, opened 2022-05-27
Final polish for the Pose Detection and some examples

zaleslaw:


zaleslaw:
@juliabeliaeva it's ready, please run the TC

onnx/src/main/kotlin/org/jetbrains/kotlinx/dl/api/inference/onnx/posedetection/MultiPoseDetectionModel.kt
```diff
@@ -38,14 +38,13 @@ public class MultiPoseDetectionModel : OnnxInferenceModel() {
         rawPoseLandMarks.forEachIndexed { poseIndex, floats ->
             val foundPoseLandmarks = mutableListOf<PoseLandmark>()
 
-            for (i in 0..50 step 3) {
+            for (keyPointIdx in 0..16) {
```
juliabeliaeva, line 41: Minor improvement: perhaps, `keyPoints` could be a list instead of a map, so we could use `keyPoints.indices` here and avoid a `!!` assertion when accessing it later.

juliabeliaeva (approved):

Issue #384, opened 2022-05-26
Error during image loading


ermolenkodev:
An exception is thrown during the loading of the jpeg image. Maybe the reason is non-standard encoding.
```
java.lang.IllegalArgumentException: Numbers of source Raster bands and source color space components do not match
```
snippet to reproduce an issue
```
val cocoPath = "000000163682.jpg"
val img = File(cocoPath).inputStream().use { inputStream -> ImageConverter.toBufferedImage(inputStream) }
```

[000000163682](https://user-images.githubusercontent.com/26741087/170472592-25d462b6-8f4e-436a-a243-0f36117fff45.jpg)

PR #404, opened 2022-07-12
Fix loading of jpeg files not supported by standard java ImageIO (#384)

ermolenkodev:
* Add TwelveMonkeys ImageIO plugin

* Add IO test with images from COCO 2017 test set, which failed to load using standard java ImageIO


ermolenkodev:
I am a bit unsure if it is a good idea to introduce TwelveMonkeys ImageIO dependency for the KotlinDL. Maybe it's better to ask users to add it on the application side.

juliabeliaeva (approved):
I'm also a bit unsure about the dependency, but we can add it now and remove it later if we don't like it. But I'd rather not add 70 images to the repository. Maybe adding just one or two test image would be enough for testing purposes?

ermolenkodev:
Yes, fair enough. I will remove images with an identical encoding issues

PR #408, opened 2022-07-26
Simplify api for building Layers

juliabeliaeva:
This pull request attempts to simplify Layer building api by doing the following:

1. Unify functional and sequential functions signatures.
2. Remove separate output shape computation (btw, it was computed incorrectly in `ZeroPadding1D` and `Concatenate` layers).
3. Combine `Layer#build` and `Layer#forward` functions.
4. Combine `GraphTrainableModel#build` and `GraphTrainableModel#forward` functions.

This change was proposed as [a part of Layer api simplification](https://github.com/Kotlin/kotlindl/issues/233#:~:text=Combine%20computeOutputShape%2C%20build%20and%20forward). Apart from the benefits described in linked proposal, having a single `build` method will make it easier to split `Layer` into an immutable `Layer` containing variables and layer output and a reusable `LayerBuilder` for building the `Layer`.
Possible downsides so far consists of one thing: the api no longer will mimic Keras api and it would be more difficult for contributors to borrow its code to implement new layers.

### DIFF
--- a/api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/Functional.kt
+++ b/api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/Functional.kt
@@ -256,7 +256,7 @@ public class Functional(vararg layers: Layer) : GraphTrainableModel(*layers) {
 
     override fun buildLayers() {
         inputLayer.build(tf)
-        inputLayer.computeOutputShape()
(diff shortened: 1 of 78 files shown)


ermolenkodev (approved):
Looks great to me! Excellent that it's not needed to calculate output shape manually anymore.

PR #418, opened 2022-08-08
Cleanups in gradle build files

juliabeliaeva:


ermolenkodev (approved):
Looks good to me! But to be honest, I don't have much experience with Gradle scripting, so I can not check for any nuances

PR #423, opened 2022-08-10
Migrate to multiplatform and publish android libraries

### LINKED ISSUE #422: Package "onnx" module as an android library

juliabeliaeva:
This PR migrates "dataset", "api" and "onnx" modules to [Kotlin Multiplatform](https://kotlinlang.org/docs/multiplatform.html) and adds android builds for them.

1. The code inside of the "dataset", "api" and "onnx" modules is split into several source sets: "commonMain", "jvmMain"/"Test" and "androidMain" for "onnx" which is currently empty. The code in the "common" source sets is included into both "jvm" and "android" libraries variants.
2. Currently, our common code is not truly platform-agnostic due to dependencies on `java.lang.AutoCloseable` and `java.io.File`. But since we only have "jvm" and "android" targets, multiplatform allows to use these classes in the common code. (see [Hierarchical project structure](https://kotlinlang.org/docs/multiplatform-hierarchy.html)). However, this PR replaced usages of `javaClass` and `System.arraycopy`, just because it was easy to do. To check which usages are platform-agnostic, add `kotlin.mpp.enableCompatibilityMetadataVariant=true` to "gradle.properties" and try to compile the project. Compilation of any platform-specific code in the "common" modules will fail.
3. A lot of our code ended up in the "jvm" source sets because of the `java.awt.image.BufferedImage` usage (not available on android). Most of the prerocessing code uses it, so the code dependent on preprocessing ended up there. A refactoring would be needed to make some of this code more platform-independent and extract to the common source sets.
4. Apart from the "jvm" and "android" publication, there is a third "multiplatform" one which represents the whole library and allows to resolve the correct library variant using included metadata (".module" file). Which means that someone using KotlinDL could just add `implementation 'org.jetbrains.kotlinx:kotlin-deeplearning-api:0.5.0'` as a dependency and it will resolve to a correct variant ("jvm" or "android"). For clients which do not support this metadata, a hack borrowed from "kotlin.serialization" is used, which adds dependency on the "jvm" variant. Not sure, how useful this is.
5. Currently, android artifacts are named incorrectly (without the "kotlin-deeplearning-" prefix). This seems to be a problem in the multiplatform, reported as [KT-53520](https://youtrack.jetbrains.com/issue/KT-53520). In any case, if the solution isn't found, we can just rename the folders (ie from "onnx" to "kotlin-deeplearning-onnx" etc).

Testing of this branch could be done by publishing artifacts locally (using "publishToMavenLocal" gradle task). After this adding `mavenLocal()` to a list of repositories and adding `implementation 'org.jetbrains.kotlinx:kotlin-deeplearning-onnx:0.5.SNAPSHOT'` dependency should work (for both "jvm" and "android").

Fixes #422


juliabeliaeva:
Merged manually in order to have a commit which moves files around separate from other changes.

PR #424, opened 2022-08-15
Introduce an interface for loading and preprocessing data

juliabeliaeva:
This PR introduces a `DataLoader` interface which represents the abstraction for loading data in a form of `FloatArray` with a `TensorShape` from some source, e.g. a file. Currently this is done in one step: loading and preprocessing together and there is no access to the intermediate data (e.g. `BufferedImage`). So far this seems OK to me, but we may want to change it in the future. Currently the interface is mainly used in the datasets, but I intend to use it for onnx models to make them more platform-independent (we should have another implementation which uses `Bitmap` as a data source for android).

After these changes `OnFlyImageDataset` needs a new name since it can work with any data now, not just images, and also "on fly" is not grammatically correct (should be "on the fly", and it also seems more like an adverb, not an adjective). My current candidates are `LazyDataset`, `LoadingDataset` and `LightweightDataset`.

Since `DataLoader` uses `TensorShape`, the latter had to be moved to the "dataset" module and it does not look too good. Perhaps, "api" module should not depend on the "dataset" and it should be the other way around.

### DIFF
--- a/api/src/jvmMain/kotlin/org/jetbrains/kotlinx/dl/api/inference/imagerecognition/ImageRecognitionModel.kt
+++ b/api/src/jvmMain/kotlin/org/jetbrains/kotlinx/dl/api/inference/imagerecognition/ImageRecognitionModel.kt
@@ -79,7 +79,7 @@ public class ImageRecognitionModel(
      *
      * @return The list of pairs <label, probability> sorted from the most probable to the lowest probable.
      */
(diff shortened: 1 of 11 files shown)


juliabeliaeva:
I removed `Preprocessing` implementing `DataLoader<File>` from this branch so that not to interfere with the #425. We can add it back later when the work in #425 would be finished. Right now I use `dataLoader()` extension function to convert `Preprocessing` to a `DataLoader` which seems to be enough at the moment.

ermolenkodev (approved):

PR #426, opened 2022-08-22
Refactorings related to the onnx high-level models

juliabeliaeva:
1. `InferenceModel` class is converted to an interface.
2. Instead of inheriting from `OnnxInferenceModel` type, high-level models delegate to an internal model provided in the constructor.
3. `OnnxInferenceModel` instances are constructed in the `ONNXModelHub`, which is a step towards having platform-specific model hub instances and using them with the same model types.
4. Minor fixes: implemented `copy` for high-level models and reduced visibility of the `OnnxInferenceModel#inputShape` property.


api/src/commonMain/kotlin/org/jetbrains/kotlinx/dl/api/inference/InferenceModel.kt
```diff
@@ -9,22 +9,22 @@ import org.jetbrains.kotlinx.dl.api.core.metric.Metrics
...
  */
-public abstract class InferenceModel : AutoCloseable {
+public interface InferenceModel : AutoCloseable {
     /** Input specification for this model. */
-    public abstract val inputDimensions: LongArray
+    public val inputDimensions: LongArray
 
     /** Model name. */
-    public var name: String? = null
+    public var name: String?
```
ermolenkodev, line 19: Why do we need the 'name' field?

examples/src/main/kotlin/examples/onnx/objectdetection/efficientdet/EfficientDetD2.kt
```diff
@@ -28,8 +28,8 @@ fun main() {
         val preprocessing: Preprocessing = preprocess {
             transformImage {
                 resize {
-                    outputHeight = it.inputShape[1].toInt()
-                    outputWidth = it.inputShape[2].toInt()
+                    outputHeight = it.inputDimensions[0].toInt()
```
ermolenkodev, line 31: In the future, we could consider supporting batch input because it's quite a popular approach to increasing throughput when users need to inference model on a full dataset

ermolenkodev (approved):

api/src/commonMain/kotlin/org/jetbrains/kotlinx/dl/api/inference/InferenceModel.kt
juliabeliaeva: I'm not sure. I guess, the name field was added because in keras models have a name. But we probably don't need it in the interface, so I'll remove it.

PR #429, opened 2022-08-26
Convert preprocessInput functions to Operations and allow to call them from the preprocessing pipeline

juliabeliaeva:
This PR aims at removing custom `preprocessInput` functions which had to be called after normal preprocessing and converting them into `Operation`s and providing api to call these in the preprocessing pipeline.


ermolenkodev (approved):
Just for discussion. It can be not so obvious that the fileLoader created from Preprocessing will combine loading and preprocessing. We could consider splitting loading and preprocessing into two separate steps.

juliabeliaeva:
> Just for discussion. It can be not so obvious that the fileLoader created from Preprocessing will combine loading and preprocessing. We could consider splitting loading and preprocessing into two separate steps.

Agree. And the syntax ".fileLoader" combined with the preprocessing pipeline looks a bit weird.

Issue #434, opened 2022-08-30
Add support of .inferUsing API for ONNX object detection and other types of models


ermolenkodev:
Add support for the following syntax

```kotlin
val model = SSDMobileNetV1ObjectDetectionModel()
val objects: List<DetectedObject> = model.inferUsing(CPU()) {
    it.detectObjects(preprocessedData, 1)
}
```

PR #446, opened 2022-09-07
Add support for SSD model on android

### LINKED ISSUE #440: Add SSD mobilenet to Android onnx models zoo
Add SSD mobilenet to android models zoo

ermolenkodev:
Fixes #440 and #434    
```
Add SSD model for android (#440)

    * Add onnxruntime-mobile dependency for androidMain

    * Reduce code duplication for SSD models code on JVM and Android platforms

    * Add support for .inferUsing API for OnnxHighLevelModel (#434)

    * Add support for zero indexed COCO labels
 ```

### DIFF
--- a/onnx/src/androidMain/kotlin/org/jetbrains/kotlinx/dl/api/inference/onnx/objectdetection/SSDMobileNetObjectDetectionModel.kt
+++ b/onnx/src/androidMain/kotlin/org/jetbrains/kotlinx/dl/api/inference/onnx/objectdetection/SSDMobileNetObjectDetectionModel.kt
@@ -1,0 +1,61 @@
+package org.jetbrains.kotlinx.dl.api.inference.onnx.objectdetection
+
+import android.graphics.Bitmap
+import org.jetbrains.kotlinx.dl.api.inference.InferenceModel
+import org.jetbrains.kotlinx.dl.api.inference.onnx.OnnxInferenceModel
+import org.jetbrains.kotlinx.dl.api.inference.onnx.executionproviders.ExecutionProvider.CPU
+import org.jetbrains.kotlinx.dl.dataset.Coco
+import org.jetbrains.kotlinx.dl.dataset.CocoVersion.V2017
+import org.jetbrains.kotlinx.dl.dataset.preprocessing.*
+import org.jetbrains.kotlinx.dl.dataset.shape.TensorShape
+
+
+private val SSD_MOBILENET_METADATA = SSDModelMetadata(
+    "TFLite_Detection_PostProcess",
+    "TFLite_Detection_PostProcess:1",
+    "TFLite_Detection_PostProcess:2",
+    0, 1
+)
+
+
+public class SSDMobileNetObjectDetectionModel(override val internalModel: OnnxInferenceModel) :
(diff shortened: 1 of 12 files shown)

### CODE REVIEW

File: onnx/src/androidMain/kotlin/org/jetbrains/kotlinx/dl/api/inference/onnx/objectdetection/SSDMobileNetObjectDetectionModel.kt [hunk 1]
```diff
@@ -0,0 +1,61 @@
...
+
+        this.targetRotation = targetRotation
+        preprocessing = buildPreprocessingPipeline()
+    }
+
+    private fun buildPreprocessingPipeline(): Operation<Bitmap, Pair<FloatArray, TensorShape>> {
+        return pipeline<Bitmap>()
+            .resize {
+                outputHeight = inputDimensions[0].toInt()
+                outputWidth = inputDimensions[1].toInt()
```
juliabeliaeva, line 72: But `inputDimensions` also changes when user runs `reshape`. That's why preprocessing in other models have a getter for preprocessing. If we want to save preprocessing to a variable here for performance reasons, we should also rebuild it when `reshape` is called.

onnx/src/androidMain/kotlin/org/jetbrains/kotlinx/dl/api/inference/onnx/objectdetection/SSDMobileNetObjectDetectionModel.kt
```diff
@@ -0,0 +1,85 @@
...
+
+    public constructor (modelBytes: ByteArray) : this(OnnxInferenceModel(modelBytes)) {
+        internalModel.initializeWith(CPU())
+        preprocessing = buildPreprocessingPipeline()
+    }
+
+    // TODO remove code duplication due to different type of class labels array
+    override fun convert(output: Map<String, Any>): List<DetectedObject> {
+        val boxes = (output[metadata.outputBoxesName] as Array<Array<FloatArray>>)[0]
+        val classIndices = (output[metadata.outputClassesName] as Array<FloatArray>)[0]
```
juliabeliaeva, line 41: Could we figure out which type it is and do a conversion? (maybe not in this pr)

onnx/src/androidMain/kotlin/org/jetbrains/kotlinx/dl/api/inference/onnx/objectdetection/SSDMobileNetObjectDetectionModel.kt
ermolenkodev: I removed the Inference model implementation, so now there is no .reshape method. Also, this model has a fixed input size, so this method didn't make sense

onnx/src/androidMain/kotlin/org/jetbrains/kotlinx/dl/api/inference/onnx/objectdetection/SSDMobileNetObjectDetectionModel.kt
ermolenkodev: Yes, it's possible to get type information using ONNX Api but it will require some work. I will add a task for it.

juliabeliaeva (approved):

Issue #438, opened 2022-09-02
Add classification models to Android onnx models zoo


ermolenkodev:
Add an initial set of mobile-friendly classification models for an Android ONNX models zoo.

PR #448, opened 2022-09-14
Add classification models for Android (#438)

ermolenkodev:
* Add EfficientNet-Lite model

* Add Mobilenetv1 model

* Implement models loading using ModelHub API

* Add Imagenet labels utils


dataset/src/commonMain/kotlin/org/jetbrains/kotlinx/dl/dataset/ImagenetUtils.kt
```diff
@@ -0,0 +1,1040 @@
...
+        labelsWithBackground[0] = "background"
+        labels.forEach { (key, value) ->
+            labelsWithBackground[key + 1] = value
+        }
+        return labelsWithBackground
+    }
+}
+
+
+public val imagenetLabels: Map<Int, String> = mapOf(
```
juliabeliaeva, line 39: These are the same labels as loaded by `org.jetbrains.kotlinx.dl.api.inference.imagerecognition.ImageRecognitionModelKt.loadImageNetClassLabels`, right? Perhaps we should get rid of the function and use this enum everywhere?

juliabeliaeva (approved):

juliabeliaeva (approved):

PR #456, opened 2022-09-26
Remove targetRotation setter api (#454)

ermolenkodev:
* Extend the ModelHub's models with an API that accepts ImageProxy as an input.

* Now user not required to set targetRotation property by hand

The rotation operation is quite expensive, so it's beneficial to make a rotation after resizing, although it makes API less intuitive.

### DIFF
--- a/onnx/src/androidMain/kotlin/org/jetbrains/kotlinx/dl/api/inference/onnx/ONNXModels.kt
+++ b/onnx/src/androidMain/kotlin/org/jetbrains/kotlinx/dl/api/inference/onnx/ONNXModels.kt
@@ -89,6 +89,7 @@ public object ONNXModels {
                 return ImageRecognitionModel(
                     modelHub.loadModel(this),
                     channelsFirst,
+                    preprocessor,
                     classLabels = Imagenet.V1001.labels()
                 )
             }
--- a/onnx/src/androidMain/kotlin/org/jetbrains/kotlinx/dl/api/inference/onnx/classification/ImageRecognitionModel.kt
+++ b/onnx/src/androidMain/kotlin/org/jetbrains/kotlinx/dl/api/inference/onnx/classification/ImageRecognitionModel.kt
@@ -1,6 +1,7 @@
 package org.jetbrains.kotlinx.dl.api.inference.onnx.classification
 
 import android.graphics.Bitmap
+import androidx.camera.core.ImageProxy
 import org.jetbrains.kotlinx.dl.api.inference.imagerecognition.ImageRecognitionModelBase
 import org.jetbrains.kotlinx.dl.api.inference.onnx.CameraXCompatibleModel
 import org.jetbrains.kotlinx.dl.api.inference.onnx.ExecutionProviderCompatible
@@ -8,6 +9,7 @@ import org.jetbrains.kotlinx.dl.api.inference.onnx.OnnxInferenceModel
 import org.jetbrains.kotlinx.dl.api.inference.onnx.executionproviders.ExecutionProvider
 import org.jetbrains.kotlinx.dl.dataset.Imagenet
 import org.jetbrains.kotlinx.dl.dataset.preprocessing.*
+import org.jetbrains.kotlinx.dl.dataset.preprocessing.camerax.toBitmap
 import org.jetbrains.kotlinx.dl.dataset.shape.TensorShape
 
 /**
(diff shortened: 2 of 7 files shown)


dataset/src/androidMain/kotlin/org/jetbrains/kotlinx/dl/dataset/preprocessing/imageproxy/imageUtlls.kt
```diff
@@ -0,0 +1,153 @@
...
+import android.graphics.Rect
+import android.graphics.BitmapFactory
+import androidx.camera.core.ImageProxy
+import java.io.ByteArrayOutputStream
+
+/**
+ * Converts an [ImageProxy] to a [Bitmap].
+ * Currently only supports [ImageFormat.YUV_420_888].
+ */
+public fun ImageProxy.toBitmap(): Bitmap? {
```
juliabeliaeva, line 15: This is a very useful function, let's merge it as a separate PR?

juliabeliaeva (changes requested):

onnx/src/androidMain/kotlin/org/jetbrains/kotlinx/dl/api/inference/onnx/ONNXModels.kt
```diff
@@ -75,13 +76,46 @@ public object ONNXModels {
...
+        override fun withImageProxyInput(modelHub: ModelHub): ImageRecognitionModel<ImageProxy> {
+            val internalModel = modelHub.loadModel(this) as OnnxInferenceModel
+
+            val (width, height) = if (channelsFirst)
+                Pair(internalModel.inputDimensions[1], internalModel.inputDimensions[2])
+            else
+                Pair(internalModel.inputDimensions[0], internalModel.inputDimensions[1])
+
+            val preprocessing = pipeline<ImageProxy>()
+                .convertToBitmap {  }
```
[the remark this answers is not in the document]
ermolenkodev: Thanks for the feedback! I agree it's a better way. Regarding performance, I hope we will address it later.

onnx/src/androidMain/kotlin/org/jetbrains/kotlinx/dl/api/inference/onnx/objectdetection/SSDLikeModel.kt
```diff
@@ -24,18 +27,33 @@ public class SSDLikeModel(override val internalModel: OnnxInferenceModel, metada
...
+/**
+ * Returns the detected object for the given image sorted by the score.
+ * Internal preprocessing is updated to rotate image to match target orientation.
+ * After prediction, internal preprocessing is restored to the original state.
+ *
+ * @param [imageProxy] Input image.
+ * @param [topK] The number of the detected objects with the highest score to be returned.
+ * @return List of [DetectedObject] sorted by score.
+ */
+public fun SSDLikeModel.detectObjects(imageProxy: ImageProxy, topK: Int = 3): List<DetectedObject> {
```
juliabeliaeva, line 55: We can write something like this to enable this method for all object detection models:

```kotlin
public fun ObjectDetectionModelBase<Bitmap>.detectObjects(imageProxy: ImageProxy, topK: Int = 3): List<DetectedObject> {
    if (this is CameraXCompatibleModel) {
        val currentRotation = targetRotation
        targetRotation = imageProxy.imageInfo.rotationDegrees
        return detectObjects(imageProxy.toBitmap(), topK).also { targetRotation = currentRotation }
    }
    return detectObjects(imageProxy.toBitmap(applyRotation = true), topK)
}
```
Not pretty, but gets the job done. What do you think?

onnx/src/androidMain/kotlin/org/jetbrains/kotlinx/dl/api/inference/onnx/classification/ImageRecognitionModel.kt
```diff
@@ -44,3 +46,37 @@ public open class ImageRecognitionModel(
...
+ * @param [imageProxy] Input image.
+ * @param [topK] Number of top ranked predictions to return
+ *
+ * @return The list of pairs <label, probability> sorted from the most probable to the lowest probable.
+ */
+public fun ImageRecognitionModel.predictTopKObjects(
+    imageProxy: ImageProxy,
+    topK: Int = 5
+): List<Pair<String, Float>> {
+    val currentRotation = targetRotation
```
juliabeliaeva, line 79: Seems like a common pattern. Do you think it would make sense to have a function like this
```kotlin
public fun <R> CameraXCompatibleModel.doWithRotation(rotation: Int, function: () -> R): R {
    val currentRotation = targetRotation
    targetRotation = rotation
    return function().apply { targetRotation = currentRotation }
}
```
?

juliabeliaeva (approved):

onnx/src/androidMain/kotlin/org/jetbrains/kotlinx/dl/api/inference/onnx/classification/ImageRecognitionModel.kt
```diff
@@ -44,3 +46,37 @@ public open class ImageRecognitionModel(
...
+    targetRotation = imageProxy.imageInfo.rotationDegrees
+    return predictTopKObjects(imageProxy.toBitmap(applyRotation = true), topK).also { targetRotation = currentRotation }
```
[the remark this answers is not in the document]
ermolenkodev: yes, sure, it's a mistake! Thanks)

Issue #211, opened 2021-09-02
Add input/outputShape information to all KDocs of layers


zaleslaw:
We need to have better documentation for our layers.
A few layers have inputShape/outputShape information in the KDocs, copied from Keras documentation, for example 
[GlobalAvgPool2D](https://github.com/JetBrains/KotlinDL/blob/41f0b20d3bfd6e277d8486d945cdef59a6250b9b/api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/pooling/GlobalAvgPool2D.kt#L20)

Please add the same documentation section to all the following layers:

- [ ] ELU
- [ ] LeakyReLU
- [ ] PReLU
- [ ] ReLU
- [ ] Softmax
- [ ] ThresholdedReLU
- [ ] DepthwiseConv2D
- [ ] SeparableConv2D
- [ ] Dense
- [ ] Input
- [ ] BatchNorm
- [ ] AvgPool1D
- [ ] AvgPool2D (should be in one format)
- [ ] AvgPool3D
- [ ] GlobalMaxPool1D
- [ ] GlobalMaxPool2D
- [ ] GlobalMaxPool3D
- [ ] MaxPool1D
- [ ] MaxPool2D
- [ ] MaxPool3D
- [ ] Flatten
- [ ] ZeroPadding1D
- [ ] ZeroPadding2D
- [ ] ZeroPadding3D

dosier:
Hey, I'd like to contribute a bit again. Could do this one. 

One question I have, in some cases Keras specifies different shapes depending upon the `data_format` being `channels_first` or `channels_last`.  I do see we have this implemented for the `ZeroPadding..` layers but not for any others. 

From what I understand the default implementation uses the `channels_last` format.

> There are two conventions for shapes of images tensors: the channels-last convention (used by TensorFlow) > and the channels-first convention (used by Theano).

 **Deep Learning with Python - François Chollet**
 (https://stackoverflow.com/a/54967450/12089702)

So I guess for now we just omit the part about the `data_format` and only take the `channels_last` part?

PR #460, opened 2022-09-29
Add input/outputShape information to all KDocs of layers (#211)

dosier:


tensorflow/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/core/Dense.kt
```diff
@@ -31,6 +31,13 @@ import org.tensorflow.op.Ops
  * created by the layer, and `bias` is a bias vector created by the layer
  * (only applicable if `use_bias` is `True`).
  *
+ * __Input shape:__  N-D tensor with shape `(batch_size, ..., input_dim)`.
```
ermolenkodev, line 34: Currently, only 2d input is supported.

tensorflow/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/reshaping/ZeroPadding1D.kt
```diff
@@ -10,6 +10,11 @@ import org.tensorflow.Shape
 /**
  * Zero-padding layer for 1D input (e.g. audio).
  * This layer can add zeros in the rows of the audio tensor
+ *
+ * __Input shape:__  3D tensor with shape `(batch_size, axis_to_pad, features)`.
```
ermolenkodev, line 14: `axis_to_pad` is a somewhat misleading name because an axis usually refers to the dimension's index, but here it refers to the size of the dimension. Just `dim` would be ok.

tensorflow/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/reshaping/ZeroPadding3D.kt
```diff
@@ -10,6 +10,11 @@ import org.tensorflow.Shape
 /**
  * Zero-padding layer for 3D input (e.g. video).
  * This layer can add zeros in the rows, cols and depth of a video tensor.
+ *
+ * __Input shape:__  5D tensor with shape `(batch_size, first_axis_to_pad, second_axis_to_pad, third_axis_to_pad, depth)`.
```
ermolenkodev, line 14: Again, maybe it would be better to replace `axis`  with  `dim`

ermolenkodev (changes requested):
Sorry for such a late review.
There are a few minor issues with the docs.
 You make the following comment in many places:
```
Use the keyword argument 'input_shape'
 (tuple of integers, does not include the samples axis)
 when using this layer as the first layer in a model.
```
But currently, we only support `Input` to be the first layer of the network. Moreover, there is no argument named `input_shape` in the constructor anywhere in the code.

dosier:
> Sorry for such a late review. There are a few minor issues with the docs. You make the following comment in many places:
> 
> ```
> Use the keyword argument 'input_shape'
>  (tuple of integers, does not include the samples axis)
>  when using this layer as the first layer in a model.
> ```
> 
> But currently, we only support `Input` to be the first layer of the network. Moreover, there is no argument named `input_shape` in the constructor anywhere in the code.

I see, thank you for the feedback! Also my apologies for the late reply, I have been quite busy.
Do u have any suggestions on what to replace that with? E.g. "Depends on the `Input` layer"'? Or can it be omitted?

tensorflow/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/layer/core/Dense.kt
dosier: Resolved in the following commit:
https://github.com/Kotlin/kotlindl/pull/460/commits/34c394f3a6df776eddc8c1bb443aed8cb0ec5356

Issue #385, opened 2022-05-26
Add validation of input shape during onnx model inference


ermolenkodev:
Currently, if an onnx model requires a 3-channel image but the input provided has only one channel internal onnx library exception will be thrown.

ai.onnxruntime.OrtException: Shape [1, 512, 512, 3], requires 786432 elements but the buffer has 262144 elements.

PR #467, opened 2022-10-11
Add validation of input shape (#385)

ermolenkodev:
* Check if the number of elements in the input matches the model's input shape

* Trying to guess if the user used a grayscale image when 3-channels expected

### DIFF
--- a/onnx/src/commonMain/kotlin/org/jetbrains/kotlinx/dl/api/inference/onnx/OnnxInferenceModel.kt
+++ b/onnx/src/commonMain/kotlin/org/jetbrains/kotlinx/dl/api/inference/onnx/OnnxInferenceModel.kt
@@ -280,6 +280,8 @@ public open class OnnxInferenceModel private constructor(private val modelSource
                                                 dataType: OnnxJavaType,
                                                 shape: LongArray
         ): OnnxTensor {
+            checkTensorMatchesInputShape(data, shape)
+
             val inputTensor = when (dataType) {
                 OnnxJavaType.FLOAT -> OnnxTensor.createTensor(this, FloatBuffer.wrap(data), shape)
                 OnnxJavaType.DOUBLE -> OnnxTensor.createTensor(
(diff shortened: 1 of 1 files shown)


onnx/src/commonMain/kotlin/org/jetbrains/kotlinx/dl/api/inference/onnx/OnnxInferenceModel.kt
```diff
@@ -320,6 +322,40 @@ public open class OnnxInferenceModel private constructor(private val modelSource
             return inputTensor
         }
 
+        private fun checkTensorMatchesInputShape(data: FloatArray, inputShape: LongArray) {
+            val numOfElements = inputShape.reduce { acc, dim -> acc * dim }.toInt()
+
+            if (data.size == numOfElements) return
+
+            if (inputShape.matchNHWC(C = 3) || inputShape.matchNCHW(C = 3)) {
+                val (height, width) = when {
```
juliabeliaeva, line 331: Do we really need to know `width` and `height` and the layout here? We can simplify this condition as something like:
```kotlin
inputShape.size == 4 && inputShape[0] == 1L && 
        (inputShape[1] == 3L || inputShape[3] == 3L) &&
                (data.size * 3 == numOfElements)
```

juliabeliaeva (approved):

onnx/src/commonMain/kotlin/org/jetbrains/kotlinx/dl/api/inference/onnx/OnnxInferenceModel.kt
ermolenkodev: agree!

onnx/src/commonMain/kotlin/org/jetbrains/kotlinx/dl/api/inference/onnx/OnnxInferenceModel.kt
```diff
@@ -320,6 +322,37 @@ public open class OnnxInferenceModel private constructor(private val modelSource
...
+                )
+            }
+
+            throw IllegalArgumentException(
+                "The number of elements (N=${data.size}) in the input tensor does not match the model input shape - "
+                    .plus("${inputShape.contentToString()}.")
+            )
+        }
+
+        private fun LongArray.matchNHWC(C: Int): Boolean {
```
juliabeliaeva, line 348: These 2 methods probably not needed anymore?

juliabeliaeva (approved):

PR #469, opened 2022-10-14
Rework module structure and packages

juliabeliaeva:
Current module structure has some problems:
1. "dataset" module contains two separate groups of classes: preprocessing api and implementation and dataset api and implementation.
2. "dataset" is not an essential module, and "api" should not have a dependency on it.
3. Some api classes, like `TensorShape` or `Operation`, would fit better in the "api" module and not in "dataset" module.

To fix these problems, this PR introduces new module structure:
1. "api" has only api classes (including preprocessing api), and no platform-specific code.
2. "impl" contains implementations, like `ImageRecognitionModelBase`, preprocessing implementations (including platform-dependent), utility classes, etc. It depends on "api" module.
3. "dataset" has dataset api and implementations. Depends on "api" and "impl" modules.

This PR also changes some packages, including moving "onnx" module classes to the package `org.jetbrains.kotlinx.dl.onnx`, and `ModelHub` and related classes from `keras.loaders` to just `loaders` (maybe here other package name like `modelhub` or `hub` could be better?). Packages in "tensorflow" module are unchanged, but we may need to refactor them in the future releases.


ermolenkodev:
It looks like some tests are not updated, at least these two:
```
impl/src/jvmTest/kotlin/org/jetbrains/kotlinx/dl/impl/preprocessing/PreprocessingTensorTest.kt
tensorflow/src/test/kotlin/org/jetbrains/kotlinx/dl/api/core/integration/SavedModelTest.kt
```

ermolenkodev:
Not related to the changes, but in this version, I cannot build the project using the command line `./gradlew,` only in IntelliJ Idea. Do you know how to use the same toolchain as in IDE when building through the command line?

ermolenkodev:
Can you please elaborate on what should go to the `impl` package in the new structure? Implementation of interfaces defined in any other project?

juliabeliaeva:
> Not related to the changes, but in this version, I cannot build the project using the command line ./gradlew, only in IntelliJ Idea. Do you know how to use the same toolchain as in IDE when building through the command line?

No, that was my mistake. I fixed it and it should work now, thank you for pointing it out.

juliabeliaeva:
> Can you please elaborate on what should go to the impl package in the new structure? Implementation of interfaces defined in any other package?

Implementations of interfaces from the "api" package, platform-specific code, additional utilities such as `CocoUtils` or `ImageNetPreprocessing`, etc.

Initially, I wanted to extract "preprocessing" module, but I noticed that a lot of classes also do not fit into the "api" category and also should be moved into a separate module.

ermolenkodev (approved):

PR #474, opened 2022-10-20
Gradle download task

### LINKED ISSUE #444: Add Gradle task for downloading models checkpoints during the build
Add Gradle task for downloading models checkpoints during the build

juliabeliaeva:
This PR adds a task `downloadKotlinDLModels` to download models from the model hub.

Plugin configuration looks like this:
```groovy
downloadKotlinDLModels {
    models = ["MoveNetSinglePoseLighting"] // list of model type names to download
    sourceSet = "main"                     // optional name of the target source set ("main" by default)
    overwrite = false                      // optional parameter to overwrite existing files ("true" by default)
}
```
Currently the plugin expects the build to have an android plugin. This is used to find the correct source set where to put the model. Since there could be several resource directories in the source set, the first one is selected. Otherwise `src/main/res/raw` directory is used.
Internally, the plugin creates a `Download` task from the [gradle-download-task](https://github.com/michel-kraemer/gradle-download-task) plugin.
Models are defined explicitly in the enum `ModelType`. In the future we could attempt to generate this enum automatically.

Fixes #444


gradlePlugin/src/main/kotlin/org/jetbrains/kotlinx/dl/gradle/ModelType.kt
```diff
@@ -0,0 +1,29 @@
+/*
+ * Copyright 2022 JetBrains s.r.o. and Kotlin Deep Learning project contributors. All Rights Reserved.
+ * Use of this source code is governed by the Apache 2.0 license that can be found in the LICENSE.txt file.
+ */
+
+package org.jetbrains.kotlinx.dl.gradle
+
+private const val ROOT_FOLDER = "models/onnx/mobile"
+private const val SUFFIX = "ort"
```
ermolenkodev, line 9: Maybe it's worth supporting both '.ort' and '.onnx' here

ermolenkodev (approved):

gradlePlugin/src/main/kotlin/org/jetbrains/kotlinx/dl/gradle/ModelType.kt
juliabeliaeva: Do you mean to support this task for desktop as well, or to support onnx models for android?

ermolenkodev: I am talking just about the file extension. We support both .ort and .onnx formats on android, but here it's supposed that only '.ort' is possible. On the other hand, we can modify it later

juliabeliaeva: No, you are right, let's support different extensions.

ermolenkodev (approved):

Issue #368, opened 2022-05-18
Add to ONNXInferenceModel human-readable model description


zaleslaw:
The existing toString method just prints some underlying information from the ONNX Java API.

We need to unify the input/output information with the TF module and also add info from modelHub if possible and about the specific type

PR #475, opened 2022-10-26
Unified summary printing for Tensorflow and ONNX models

ermolenkodev:
Fixes (#368)

- Add common ModelSummary interface for TensorFlow and onnx models
- Refactor summary functionality to use the new ModelSummary interface
- Add missing KDocs
- Update tests and examples


impl/src/commonMain/kotlin/org/jetbrains/kotlinx/dl/impl/inference/imagerecognition/ImageRecognitionModelBase.kt
```diff
@@ -5,17 +5,24 @@
...
 
 /**
  * Base class for image classification models.
  * @property [internalModel] model used for prediction
+ * @property [type] Corresponding [ModelType]. Used for model summary. Should be null for custom models.
  */
 public abstract class ImageRecognitionModelBase<I>(
     protected val internalModel: InferenceModel,
-) : InferenceModel by internalModel {
+    protected val type: ModelType<*, *>? = null
```
juliabeliaeva, line 24: I'd rather added something more broad here. We only need this parameter for the summary, why don't we add something like `modelKindDescription: String` or just make `type` a `String`. This will allow users with custom models use this property as well.

juliabeliaeva (approved):

ermolenkodev:
@juliabeliaeva I refactored this as you suggested, with much less copy-pasting. Could you also please rerun the tests on the CI?

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/summary/FormatUtils.kt
```diff
@@ -0,0 +1,261 @@
...
+ * Convenient function to calculate the required width of all columns.
+ * @param [rows] table with multiple rows each with multiple columns
+ * @param [columnNames] list of column names
+ *
+ * Note: each row must have the same number of columns as the [columnNames] list.
+ * Note: each row may consist of multiple lines.
+ */
+public fun calcColumnWidths(rows: List<TableRow>, columnNames: List<String> = emptyList()): MutableList<Int> {
+    if (columnNames.isEmpty()) {
+        return listOf(rows.flatMap { it.lines() }.maxOfOrNull { it.size } ?: 0).toMutableList()
```
juliabeliaeva, line 24: `calcColumnWidths(listOf(TableRow(listOf("one", "two", "three"))))` returns `[3]` for me. Is this correct?

tensorflow/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/summary/TfModelSummary.kt
```diff
@@ -0,0 +1,130 @@
...
+        paramsCountColumnName: String = "Param #      ",
+        connectedToColumnName: String = "Connected to               ",
+        columnSeparator: String = " ",
+        lineSeparatorSymbol: Char = '_',
+        thickLineSeparatorSymbol: Char = '=',
+        withConnectionsColumn: Boolean = layersSummaries.any { it.inboundLayers.size > 1 }
+    ): List<String> {
+        // Prepare string data for resulting table
+        val rows = layersSummaries.map { layer ->
+            SummaryRow(
```
juliabeliaeva, line 60: Perhaps we just should create a `TableRow` directly, without the `SummaryRow` class?

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/summary/FormatUtils.kt
```diff
@@ -0,0 +1,261 @@
...
+        columnsWidths = columnsWidths.map { it + (newWidth - getWidth()) / columnsWidths.size }.toMutableList()
+    }
+}
+
+
+/**
+ * The row of the summary table may consist of multiple lines.
+ * This class is used to simplify the handling of formatting of the rows.
+ */
+public class TableRow(lines: List<List<String>>) {
```
juliabeliaeva, line 239: I would expect that a row consists of cells, and each cell can have a multiple lines of text. This would simplify creating rows in the model code.

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/summary/FormatUtils.kt
```diff
@@ -0,0 +1,261 @@
...
+    val sectionsWithColumns = sections.filterIsInstance<SectionWithColumns>()
+
+    for (section in sectionsWithColumns) {
+        section.alignWith(sectionsWithColumns)
+    }
+
+    val simpleSections = sections.filterIsInstance<SimpleSection>()
+
+    for (section in simpleSections) {
+        section.alignWith(sectionsWithColumns)
```
juliabeliaeva, line 71: Mm, so the formatting method alters sections passed to it? This would be really unexpected to the caller, let's not do this.

Also, passing information via object fields is super-not-evident. Let's calculate the widths here explicitly, save them to the variable and then pass to the formatting method.

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/summary/FormatUtils.kt
```diff
@@ -0,0 +1,261 @@
...
+}
+
+
+/**
+ * Section with columns of the model summary table.
+ * Usually it is a header or footer of the table.
+ */
+public class SimpleSection(
+    private val rows: List<TableRow>,
+    private val lineSeparatorSymbol: Char,
```
juliabeliaeva, line 120: Do we want separator symbols as a part of a model? They seem to be more related to the presentation. So we can do something like `section.format(lineSeparatorSymbol = "*")` (passing different separators to format the same section).

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/summary/FormatUtils.kt
```diff
@@ -0,0 +1,261 @@
...
+ * Section with columns of the model summary table.
+ */
+public class SectionWithColumns(
+    private val rows: List<TableRow>,
+    private val columnNames: List<String>,
+    private val columnSeparator: String,
+    private val lineSeparatorSymbol: Char,
+    private val thickLineSeparatorSymbol: Char
+) : Section {
+    private var columnsWidths = calcColumnWidths(rows, columnNames)
```
juliabeliaeva, line 167: Does this field need to be a `MutableList` and a `var` at the same time?

api/src/main/kotlin/org/jetbrains/kotlinx/dl/api/summary/FormatUtils.kt
```diff
@@ -0,0 +1,320 @@
...
+): List<String> {
+    require(sections.isNotEmpty()) { "At least one section must be provided" }
+
+    val sectionsWithColumns = sections.filterIsInstance<SectionWithColumns>()
+    val simpleSections = sections.filterIsInstance<SimpleSection>()
+
+    val maxSimpleSectionWidth = simpleSections.maxOf(SimpleSection::getWidth)
+
+    // here we calculate the width of each column of each section so each column will be aligned
+    val sectionsAlignedColumnsWidths = buildMap {
```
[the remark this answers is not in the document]
juliabeliaeva: Something like this:
```kotlin
    require(sections.isNotEmpty()) { "At least one section must be provided" }

    val sectionsWithColumns = sections.filterIsInstance<SectionWithColumns>()
    val columnsCount = sectionsWithColumns.maxOf { it.columnsCount }
    val columnWidth = List(columnsCount) { column ->
        sectionsWithColumns.maxOfOrNull { section -> section.columnWidth(column) } ?: 0
    }

    val simpleSectionsWidth = sections.filterIsInstance<SimpleSection>().maxOfOrNull(SimpleSection::width) ?: 0
    val sectionsWithColumnsWidth = columnWidth.sum() + (columnWidth.size - 1).coerceAtLeast(0) * columnSeparator.length
    val tableWidth = max(simpleSectionsWidth, sectionsWithColumnsWidth)

    val result = mutableListOf<String>()
    for (section in sections) {
        val lines = when (section) {
            is SimpleSection -> section.format(tableWidth, lineSeparatorSymbol, thickLineSeparatorSymbol)
            is SectionWithColumns -> section.format(
                columnWidth, tableWidth,
                columnSeparator, lineSeparatorSymbol, thickLineSeparatorSymbol
            )
        }
        result.addAll(lines)
    }
    return result
```

(see https://github.com/Kotlin/kotlindl/commit/d4a90e55e8040a8ca98620edd3bc32a3ec0cc737)
(this example does not pad the last column to the full table width, but it won't be hard to add)

juliabeliaeva (changes requested):

juliabeliaeva (approved):

PR #480, opened 2022-11-09
Add a FlatShape interface to allow manipulating the detected shapes in a unified way

juliabeliaeva:
Adding a common interface allows to simplify writing code like this:

```
private fun <T : FlatShape<T>> T.flip(): T {
    return map { x, y -> 1 - x to y }
}

private fun <T : FlatShape<T>> T.flipIfNeeded(isImageFlipped: Boolean): T {
    if (isImageFlipped) return flip()
    return this
}
```

Flipping detection results horizontally is used on Android when the front camera is used.


juliabeliaeva:
At the same time, this could be done with the `when` statement, and `when` is still used when drawing detection results in the Android demo app. So I'm not sure how necessary this is. @ermolenkodev what do you think?

ermolenkodev:
I like it! I think the nice API is the main selling point of this library so this change fits perfectly

ermolenkodev (approved):

PR #494, opened 2022-12-06
Add basic multiple input support to OnnxInferenceModel

### LINKED ISSUE #417: multiple argument support for the Input in OnnxInferenceModel 's predictRaw and it's sibling
was trying to import model with 2 parameter in kDL but apparently doesn't seems to be supported because as in "https://github.com/Kotlin/kotlindl/blob/349d09d1a3817b89c12302a0d373a16b5f7e3fa3/onnx/src/main/kotlin/org/jetbrains/kotlinx/dl/api/inference/onnx/OnnxInferenceModel.kt" no matter how much argument it needs, it would only emit single input because the 3 deriving function seems to take the first element of `onnxsession.inputName` into `onnxsession.run(()` 
I there any plan to resolve this issue ?

juliabeliaeva:
Changes in this PR:
1. `FloatData` type is added to represent data with shape and to replace `Pair<FloatArray, TensorShape>`.
2. `OnnxInferenceModel` accepts `FloatData` in different prediction methods.
3. `OnnxInferenceModel` accepts multiple inputs in a form `Map<String, FloatData>`.

Methods inherited from the `InferenceModel` still accept data without shape information (`FloatArray`). This is a problem for models which work with data of variable shapes, since shape information is missing. So these models keep `inputShape` set externally with `reshape` method in a separate field. Ideally, this method should not exist, but this requires refactoring to make `InferenceModel.predict` to work with `FloatData`, thus `TensorflowInferenceModel` also needs to accept `FloatData` and all of the datasets need to work with `FloatData` as well. This refactoring is planned for later.

In the future, `FloatData` could be replaced with multik arrays.

Fixes #417


ermolenkodev (approved):
The only note that there are a lot of unused imports after refactoring
`
import org.jetbrains.kotlinx.dl.api.core.shape.TensorShape
`

PR #512, opened 2023-01-03
Cleanup TensorFlow model classes

juliabeliaeva:
This PR:
1. Removes copy-pasted code between prediction and training methods.
2. Extracts `TensorflowInferenceModelBase` as a base class for TF models. `GraphTrainableModel` does not implement `TensorflowInferenceModel` anymore.
3. Moves properties which are effectively final to constructor parameters.


tensorflow/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/util/tensorNames.kt
```diff
@@ -16,3 +16,5 @@ internal const val OUTPUT_ARG_MAX = "ArgMax"
 internal const val DATA_PLACEHOLDER = "default_data_placeholder"
 
 internal const val LABEL_PLACEHOLDER = "default_label_placeholder"
```
ermolenkodev, line 18: CLASSIFIER_OUTPUT_NAME and LABEL_PLACEHOLDER are unused

ermolenkodev (approved):
Looks good! But I am not sure if I completely understand this PR. I see a reduction in code duplication and a change in the hierarchy of classes. Are there not-so-obvious changes that affect other modules or APIs?

juliabeliaeva:
> Are there not-so-obvious changes that affect other modules or APIs?

No, there are only cleanups. I don't want to lump them with more meaningful changes.

PR #513, opened 2023-01-06
Keep dataset shape and use shaped data for training and prediction

juliabeliaeva:
This PR is a continuation of #494. It adds shape information for all the data used for prediction and training. Since `shape` is always known, we can create tensors for the data more easily.

Some models do not have the input shape defined and data with arbitrary shape can be used (ex `EfficientDetObjectDetectionModel`). In this PR such models still use `resize` operation in preprocessing, so they still have some input shape set manually for this. Maybe this is not needed (?)


ermolenkodev (approved):
> Some models do not have the input shape defined and data with arbitrary shape can be used (ex EfficientDetObjectDetectionModel). In this PR such models still use resize operation in preprocessing, so they still have some input shape set manually for this. Maybe this is not needed (?)

I think it's pretty common to explicitly resize image before inference, because size have effect on performance, moreover some models works better with specific input sizes although don't have any

PR #515, opened 2023-01-09
Introduce new general-purpose prediction methods for InferenceModel

juliabeliaeva:
This PR removes classification methods from the `InferenceModel` interface in favor of the more general api. It introduces two prediction methods: for single input and for multiple inputs and adds api to extract prediction results from the model. Instance methods are replaced with the following options for classification:
1. If a model outputs a vector of probabilities, there are `PredictionKt#predictLabel`/`PredictionKt#predictProbabilities` extensions or the model could be wrapped into `ImageRecognitionModelBase`.
2. For models with different outputs, function `InferenceModel#predict(kotlin.Pair<float[], TensorShape>, Function1<? super R,? extends T>)` could be used combined with the `InferenceResultConverter` for result extraction.

Additional comments:
1. TensorFlow models still have some methods for classification left, for convenience.
3. `TensorResult` right now only has a list of tensors, but ideally it could also contain a map "output name" -> "tensor" and allow access by `String`.
4. `InferenceResultConverter` only has 2 methods returning `LongArray` and `FloatArray`, this list should be expanded for other data types.
5. `InferenceModel` does not provide enough input-output information. Currently it only has `inputDimensions` property, instead we should replace it with something similar to `OnnxInferenceModel#inputInfo`/`outputInfo` to get access to all inputs and outputs.


ermolenkodev (approved):
I like the design. It's great that we don't have hardcode classification-related methods in interfaces anymore. Thanks!

Issue #530, opened 2023-04-09
Title of plot not fully readable


b0n541:
When I run the LeNetMnistVisualization from the examples module 

https://github.com/Kotlin/kotlindl/blob/master/examples/src/main/kotlin/examples/visualization/LeNetMnistVisualization.kt

the title of the plot is not fully displayed. One can't read the predicted label:

![image](https://user-images.githubusercontent.com/2326303/230784623-2cf17f98-c99b-4f25-8ba9-dcb994cb5de0.png)

It would be better to introduce a line break instead of the pipe symbol between the real and predicted label.

b0n541:
I would like to fix this issue.

b0n541:
This is how it would look like:

![image](https://user-images.githubusercontent.com/2326303/230788681-0ff6d438-d401-4521-893b-ae3020f60b3d.png)

juliabeliaeva:
@b0n541 looks nice, a PR would be very welcome!

PR #531, opened 2023-04-09
Fix for issue #530

b0n541:
This fixes the issue #530 by changing the title of plots.


juliabeliaeva (approved):
Thank you for the fix!

PR #534, opened 2023-04-11
Create all necessary parent directories when saving models

### LINKED ISSUE #525: Model saving is failed because Save function could not create the full path if it doesn't exist
We have a model that should be saved on the disk.
But after calling the save method, the following exception is thrown, which is not expected.

`private const val PATH_TO_MODEL = "savedmodels/customResNet50"`

```kotlin
        it.save(
            File(PATH_TO_MODEL),
            SavingFormat.JSON_CONFIG_CUSTOM_VARIABLES,
            writingMode = WritingMode.OVERRIDE
        )
```

`Exception in thread "main" java.io.FileNotFoundException: C:\.....\savedmodels\customResNet50\modelConfig.json (The system cannot find the path specified)
	at java.base/java.io.FileOutputStream.open0(Native Method)
	at java.base/java.io.FileOutputStream.open(FileOutputStream.java:298)
	at java.base/java.io.FileOutputStream.<init>(FileOutputStream.java:237)
	at java.base/java.io.FileOutputStream.<init>(FileOutputStream.java:187)
	at kotlin.io.FilesKt__FileReadWriteKt.writeBytes(FileReadWrite.kt:108)
	at kotlin.io.FilesKt__FileReadWriteKt.writeText(FileReadWrite.kt:134)
	at org.jetbrains.kotlinx.dl.api.inference.keras.ModelSaverKt.saveModelConfiguration(ModelSaver.kt:43)
	at org.jetbrains.kotlinx.dl.api.core.Functional.saveModel(Functional.kt:320)
	at org.jetbrains.kotlinx.dl.api.core.Functional.saveModel$default(Functional.kt:318)
	at org.jetbrains.kotlinx.dl.api.core.Functional.saveInKerasFormat(Functional.kt:314)
	at org.jetbrains.kotlinx.dl.api.core.Functional.save(Functional.kt:309)
	at org.jetbrains.kotlinx.dl.api.core.TrainableModel.save$default(TrainableModel.kt:386)
	at cnn.TransferLearningExportImportKt.main(transferLearningExportImport.kt:104)
	at cnn.TransferLearningExportImportKt.main(transferLearningExportImport.kt)`

P.S. It was fixed when I created the folder `savedmodels` manually

juliabeliaeva:
Fixes #525

### DIFF
--- a/tensorflow/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/GraphTrainableModel.kt
+++ b/tensorflow/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/GraphTrainableModel.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright 2020-2022 JetBrains s.r.o. and Kotlin Deep Learning project contributors. All Rights Reserved.
+ * Copyright 2020-2023 JetBrains s.r.o. and Kotlin Deep Learning project contributors. All Rights Reserved.
  * Use of this source code is governed by the Apache 2.0 license that can be found in the LICENSE.txt file.
  */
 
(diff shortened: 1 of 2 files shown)


tensorflow/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/GraphTrainableModel.kt
```diff
@@ -825,20 +825,17 @@ public abstract class GraphTrainableModel(vararg layers: Layer) : TrainableModel
         when (writingMode) {
             WritingMode.FAIL_IF_EXISTS -> {
                 check(!modelDirectory.exists()) { "The directory exists on path $pathToModelDirectory, please be careful it could contain valuable model! Change this mode to OVERRIDE if you want to override this directory." }
-                Files.createDirectories(modelDirectory.toPath())
```
zaleslaw, line 828: I need to say, that this is not the same code with replacement of createDirectories + mkdir => mkdirs due to reducing number of checks and supported exceptions

But I agree, it should be consistent in both files

zaleslaw (approved):

tensorflow/src/main/kotlin/org/jetbrains/kotlinx/dl/api/core/GraphTrainableModel.kt
zaleslaw: Wow, really, double creation of the last directory, ok, I will check and return

PR #542, opened 2023-04-24
Fix dimension order in the image shape

juliabeliaeva:
The order of the dimensions did not matter that much in the 0.5.* version, since it's rarely used. But in 0.6 we use the output shape to create tensors passed to the models, so it's important to have this correct. In quite a few models input width is equal to the input height, but for the Ultraface models these dimensions are different, so the `FaceDetectionWithVisualization.kt` example will fail in master branch without the fix.

### DIFF
--- a/impl/src/jvmTest/kotlin/org/jetbrains/kotlinx/dl/impl/preprocessing/PreprocessingFinalShapeTest.kt
+++ b/impl/src/jvmTest/kotlin/org/jetbrains/kotlinx/dl/impl/preprocessing/PreprocessingFinalShapeTest.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright 2020-2022 JetBrains s.r.o. and Kotlin Deep Learning project contributors. All Rights Reserved.
+ * Copyright 2020-2023 JetBrains s.r.o. and Kotlin Deep Learning project contributors. All Rights Reserved.
  * Use of this source code is governed by the Apache 2.0 license that can be found in the LICENSE.txt file.
  */
 
@@ -47,7 +47,7 @@ class PreprocessingFinalShapeTest {
                 bottom = 7
             }
             .toFloatArray { }
-        assertEquals(TensorShape(186, 188, 3), preprocess.getOutputShape(TensorShape(200, 200, 3)))
+        assertEquals(TensorShape(178, 186, 3), preprocess.getOutputShape(TensorShape(190, 200, 3)))
     }
 
     @Test
@@ -66,7 +66,7 @@ class PreprocessingFinalShapeTest {
                 bottom = 3
             }
             .toFloatArray { }
-        assertEquals(TensorShape(180, 180, 3), preprocess.getOutputShape(TensorShape(200, 200, 3)))
+        assertEquals(TensorShape(380, 180, 3), preprocess.getOutputShape(TensorShape(400, 200, 3)))
     }
 
     @Test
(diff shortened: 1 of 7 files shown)


impl/src/jvmTest/kotlin/org/jetbrains/kotlinx/dl/impl/preprocessing/PreprocessingFinalShapeTest.kt
```diff
@@ -107,7 +107,7 @@ class PreprocessingFinalShapeTest {
                 right = 13
             }
             .toFloatArray { }
-        assertEquals(TensorShape(324, 212, 1), preprocess.getOutputShape(TensorShape(300, 200, 1)))
+        assertEquals(TensorShape(312, 224, 1), preprocess.getOutputShape(TensorShape(300, 200, 1)))
```
zaleslaw, line 110: Why it's here 312 and 224 instead of 212 and 324?

impl/src/jvmTest/kotlin/org/jetbrains/kotlinx/dl/impl/preprocessing/PreprocessingFinalShapeTest.kt
```diff
@@ -47,7 +47,7 @@ class PreprocessingFinalShapeTest {
                 bottom = 7
             }
             .toFloatArray { }
-        assertEquals(TensorShape(186, 188, 3), preprocess.getOutputShape(TensorShape(200, 200, 3)))
+        assertEquals(TensorShape(178, 186, 3), preprocess.getOutputShape(TensorShape(190, 200, 3)))
```
zaleslaw, line 50: could you please explain these numbers changes?

impl/src/jvmTest/kotlin/org/jetbrains/kotlinx/dl/impl/preprocessing/PreprocessingFinalShapeTest.kt
```diff
@@ -66,7 +66,7 @@ class PreprocessingFinalShapeTest {
                 bottom = 3
             }
             .toFloatArray { }
-        assertEquals(TensorShape(180, 180, 3), preprocess.getOutputShape(TensorShape(200, 200, 3)))
+        assertEquals(TensorShape(380, 180, 3), preprocess.getOutputShape(TensorShape(400, 200, 3)))
```
zaleslaw, line 69: why it was increased two times for output and not twice for input?

zaleslaw (changes requested):
I am in doubt about calculations, is it correct?

impl/src/jvmTest/kotlin/org/jetbrains/kotlinx/dl/impl/preprocessing/PreprocessingFinalShapeTest.kt
juliabeliaeva: The number changed because dimension order changed. Before we used width, height, channels order, now height, width, channels. I also changed the input image height from 200 to 190 so the image would not be square. I figure it's better for testing.

impl/src/jvmTest/kotlin/org/jetbrains/kotlinx/dl/impl/preprocessing/PreprocessingFinalShapeTest.kt
juliabeliaeva: Input image height is 300, we pad it with 5 pixels on top and 7 on the bottom. So 300 + 5 + 7 = 312. Input width is 200, we pad is with 11 pixels on the left and 13 pixels on the right -- 200 + 11 + 13 = 224.

The dimensions order in the tensor shape switched, but the padding values stayed the same, so because of this it's not a simple swap of the numbers in the expected result.

juliabeliaeva:
I checked once again, everything seems correct to me.

zaleslaw (approved):

Issue #544, opened 2023-04-24
Face alignment detection with Fan2d106 only works well with a certain image types


juliabeliaeva:
The model only return good results for images with big faces at the image center, and generally needs to be combined with a face detection method to detect the face bounding box. The following things need to be done:

- [ ] Add a desktop example of combining face detection and face alignment.
- [ ] Determine parameters for "bad" and "good" images.
- [ ] Add information to documentation.

juliabeliaeva:
See #541

Issue #541, opened 2023-04-22
Detect face landmarks with Fan2d106 returns strange landmarks


lubitelfoto:
`val modelHub = ONNXModelHub(cacheDirectory = File("cache/pretrainedModels"))
    val model = modelHub.loadPretrainedModel(ONNXModels.FaceAlignment.Fan2d106)
    val landmarks = model.detectLandmarks(imageFile = imageFile)`
return
<img width="200" alt="Screenshot 2023-04-20 at 16 08 44" src="https://user-images.githubusercontent.com/15362866/233780249-70daaedc-fbc4-4b08-80ec-41dce89bc9ce.png">

`val modelHub = ONNXModelHub(cacheDirectory = File("cache/pretrainedModels"))
  val model = ONNXModels.PoseDetection.MoveNetSinglePoseLighting.pretrainedModel(modelHub)
    model.use { poseDetectionModel ->
        val detectedPoses = poseDetectionModel.detectPose(imageFile = imageFile)
        var landmarks = mutableListOf<Landmark>()
                detectedPoses.landmarks.forEach { (x,y,_,label) ->
                    if (label=="nose"||label=="left_eye"||label=="right_eye") landmarks.add(Landmark(x,y))
                }`
return

<img width="200" alt="Screenshot 2023-04-22 at 15 03 49" src="https://user-images.githubusercontent.com/15362866/233780386-e5e7207b-1847-4e88-9559-f6a03847f646.png">

Figure get the same method
`landmarks.forEach { (x, y) ->
                    points.add(Offset(this@onDraw.size.width * x , this@onDraw.size.height * y )) }
                drawPoints(points = points,
                    pointMode =  PointMode.Points,
                    color = Color.Red,
                    strokeWidth = 10f
                )`

zaleslaw:
I faced with the same issue earlier, give me a time to have a look into the output

juliabeliaeva:
Fan2d106 model works well only when the face is roughly in the image center and takes a big part of it. In other cases you need to localise the face first. You can do it with the Ultraface face detection model. This is done in the sample Android app (see here https://github.com/Kotlin/kotlindl-app-sample/blob/e82c40e91f1996798639fd0b561c1d2f00800c8a/app/src/main/java/org/jetbrains/kotlinx/dl/example/app/pipelines.kt#L199).

lubitelfoto:
[...]

PR #545, opened 2023-04-24
Add an example of combining face detection and face alignment models

juliabeliaeva:
This is a demonstration of how to overcome the problem with the face alignment model described in #541 and is a part of fixing #544. Only works after merging #542 because of the problems with the dimensions order.

Result:
image

### DIFF
--- a/examples/src/main/kotlin/examples/onnx/faces/FaceDetectionAndAlignmentWithVisualization.kt
+++ b/examples/src/main/kotlin/examples/onnx/faces/FaceDetectionAndAlignmentWithVisualization.kt
@@ -1,0 +1,80 @@
+/*
+ * Copyright 2023 JetBrains s.r.o. and Kotlin Deep Learning project contributors. All Rights Reserved.
+ * Use of this source code is governed by the Apache 2.0 license that can be found in the LICENSE.txt file.
+ */
+
+package examples.onnx.faces
+
+import examples.transferlearning.getFileFromResource
+import org.jetbrains.kotlinx.dl.api.inference.FlatShape
+import org.jetbrains.kotlinx.dl.api.inference.objectdetection.DetectedObject
+import org.jetbrains.kotlinx.dl.api.preprocessing.pipeline
+import org.jetbrains.kotlinx.dl.api.summary.printSummary
+import org.jetbrains.kotlinx.dl.impl.preprocessing.image.ImageConverter
+import org.jetbrains.kotlinx.dl.impl.preprocessing.image.crop
+import org.jetbrains.kotlinx.dl.impl.preprocessing.image.resize
+import org.jetbrains.kotlinx.dl.onnx.inference.ONNXModelHub
+import org.jetbrains.kotlinx.dl.onnx.inference.ONNXModels
+import org.jetbrains.kotlinx.dl.onnx.inference.OnnxHighLevelModel
+import org.jetbrains.kotlinx.dl.visualization.swing.createDetectedLandmarksPanel
+import org.jetbrains.kotlinx.dl.visualization.swing.showFrame
+import java.awt.image.BufferedImage
+import java.io.File
+
+/**
+ * This example demonstrates how to combine Fan2d106 face alignment model and UltraFace320 face detection model to
+ * find face landmarks on the image. The face alignment model works well only when the face has a certain size and location,
+ * so for other cases it is necessary to find the face location first with a face detection model
+ * and only then apply face alignment model to detect landmarks on the face crop.
+ */
+fun main() {
+    val image = ImageConverter.toBufferedImage(getFileFromResource("datasets/poses/multi/2.jpg"))
+
+    val modelHub = ONNXModelHub(cacheDirectory = File("cache/pretrainedModels"))
+
+    val detectionModel = ONNXModels.FaceDetection.UltraFace320.pretrainedModel(modelHub)
+    detectionModel.printSummary()
+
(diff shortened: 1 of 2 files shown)


examples/src/main/kotlin/examples/onnx/faces/FaceDetectionAndAlignmentWithVisualization.kt
```diff
@@ -0,0 +1,80 @@
...
+    val faces = detectionModel.use { it.detectFaces(image) }
+
+    val alignmentModel = ONNXModels.FaceAlignment.Fan2d106.pretrainedModel(modelHub)
+    alignmentModel.printSummary()
+
+    val facesToLandmarks = alignmentModel.use {
+        faces.associateWith { face -> alignmentModel.predictOnCrop(image, face) }
+    }
+
+    val width = 600
```
zaleslaw, line 47: Let's move to constants

examples/src/main/kotlin/examples/onnx/faces/FaceDetectionAndAlignmentWithVisualization.kt
```diff
@@ -0,0 +1,80 @@
...
+private fun <T : FlatShape<T>> OnnxHighLevelModel<BufferedImage, List<T>>.predictOnCrop(image: BufferedImage,
+                                                                                        face: DetectedObject
+): List<T> {
+    val faceWidth = face.xMax - face.xMin
+    val faceHeight = face.yMax - face.yMin
+    val x1 = ((face.xMin - 0.1f * faceWidth) * image.width).toInt().coerceAtLeast(0)
+    val y1 = ((face.yMin - 0.1f * faceHeight) * image.height).toInt().coerceAtLeast(0)
+    val x2 = ((face.xMax + 0.1f * faceWidth) * image.width).toInt().coerceAtMost(image.width)
+    val y2 = ((face.yMax + 0.1f * faceHeight) * image.height).toInt().coerceAtMost(image.height)
+    val cropImage = pipeline<BufferedImage>().crop {
```
zaleslaw, line 67: lets add a blank line

examples/src/main/kotlin/examples/onnx/faces/FaceDetectionAndAlignmentWithVisualization.kt
```diff
@@ -0,0 +1,80 @@
...
+        left = x1
+        top = y1
+        right = image.width - x2
+        bottom = image.height - y2
+
+    }.apply(image)
+    return predict(cropImage).map { shape ->
```
zaleslaw, line 74: let's add a blank line again

examples/src/main/kotlin/examples/onnx/faces/FaceDetectionAndAlignmentWithVisualization.kt
```diff
@@ -0,0 +1,80 @@
...
+        outputWidth = width
+        outputHeight = width * image.height / image.width
+    }
+    showFrame(
+        "Detected Landmarks For ${facesToLandmarks.size} Face" + (if (facesToLandmarks.size == 1) "" else "s"),
+        createDetectedLandmarksPanel(resize.apply(image), facesToLandmarks.values.flatten())
+    )
+}
+
+private fun <T : FlatShape<T>> OnnxHighLevelModel<BufferedImage, List<T>>.predictOnCrop(image: BufferedImage,
```
zaleslaw, line 58: cool function, let's add some notes about it and probably make public (give some hints to the user, if he want to copy our code)

How do you think, do we need this in our utils?

examples/src/main/kotlin/examples/onnx/faces/FaceDetectionAndAlignmentWithVisualization.kt
```diff
@@ -0,0 +1,80 @@
...
+    val image = ImageConverter.toBufferedImage(getFileFromResource("datasets/poses/multi/2.jpg"))
+
+    val modelHub = ONNXModelHub(cacheDirectory = File("cache/pretrainedModels"))
+
+    val detectionModel = ONNXModels.FaceDetection.UltraFace320.pretrainedModel(modelHub)
+    detectionModel.printSummary()
+
+    val faces = detectionModel.use { it.detectFaces(image) }
+
+    val alignmentModel = ONNXModels.FaceAlignment.Fan2d106.pretrainedModel(modelHub)
```
zaleslaw, line 40: Sorry, I don't see where is closed this model?

zaleslaw (changes requested):

examples/src/main/kotlin/examples/onnx/faces/FaceDetectionAndAlignmentWithVisualization.kt
juliabeliaeva: Yes, it seems useful, I'll move it from the examples to a better location.

juliabeliaeva:
> Also, the place is unclear—if it's not a runnable test, probably better to move into the visualization folder

It's not very different from other files in this package, like `Fan2D106LightAPIwithVisualization.kt`. Should all of the similar examples be moved?

examples/src/main/kotlin/examples/onnx/faces/FaceDetectionAndAlignmentWithVisualization.kt
zaleslaw: haha, I missed the use usage, sorry

zaleslaw (approved):

PR #566, opened 2023-08-30
Converting Gradle Groovy -> Gradle Kotlin DSL for libraries

mjovanc:
Note that this does not include the `.gradle` files in the `gradle` directory. I will convert these too at a later point. But let me know what you think.


juliabeliaeva:
Would be nice if the conversion was split into 2 commits: the first with file renames from "*.gradle" to "*.gradle.kts" without changing the content; the second with the file content conversion. This will allow to preserve history of the build files.

zaleslaw:
Thank you for the PR, I will review it and merge if it's ok during the next 2–3 weeks.

